PR #3, opened 2026-04-01
feat: enhance run-automation-suite with app selection, background execution, and browser session

d4rkwinz:
## Summary

- **Mandatory app selection prompt**: Always asks user whether to upload a new app or reuse an existing one before proceeding, preventing accidental uploads
- **Background script execution**: Runs Appium test scripts in the background so users can interact with the running session in real-time
- **Browser session integration**: New Step 5 opens the running session in the user's browser (with configurable browser preference saved to memory) via the Kobiton portal launch URL
- **Version bump to 1.0.1** with updated changelog

## Test plan

- [x] Verify the skill prompts for app selection before any upload/app tools are called
- [x] Confirm test scripts execute in the background (`run_in_background: true`)
- [x] Test browser launch opens the correct portal URL based on MCP server config
- [x] Validate version bump and changelog entries are accurate

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8, opened 2026-04-15
feat: streamline browser launch to use the default browser always.

ufx:
Ensure Claude all steps including browser launch and result collection, and streamline browser launch.

### DIFF
--- a/skills/run-automation-suite/SKILL.md
+++ b/skills/run-automation-suite/SKILL.md
@@ -94,17 +94,9 @@ Session Name: Verify Appium session
 Command:      node /path/to/test.js 9B211FFAZ0017F
 ```
 
-Wait for user confirmation, then execute the command via the Bash tool **in the background** (use `run_in_background: true`). This allows the user to review the running session in the browser while the script executes.
+Wait for user confirmation, then execute the command via the Bash tool **in the background** (use `run_in_background: true`).
 
-### 5. Open running session in browser
-
-Ask the user:
-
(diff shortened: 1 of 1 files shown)


skills/run-automation-suite/SKILL.md
```diff
@@ -121,30 +113,16 @@ If they agree, wait **2 seconds** after the script was launched in Step 4 (to al
 
 Where `<deviceId>` is the ID of the selected device from Step 2 (returned by `listDevices`, `getDeviceStatus`, or `reserveDevice`).
 
-**Browser preference:** Check auto memory for a saved browser preference. If none exists, ask the user which browser to use:
```
d4rkwinz, line 124: I have commented here https://github.com/kobiton/automate/pull/3#discussion_r3090738052. But if you feel we only need to use only the default browser. I'm fine with it.

d4rkwinz:
@ufx please help to patch the plugin version to `1.0.3` otherwise, you Claude may not update it correctly.

skills/run-automation-suite/SKILL.md
ufx: Yes, I feel we should use only the default browser, and moreover that agents should act as autonomously as possible without bothering the user with non-essential questions.

ufx:
> @ufx please help to patch the plugin version to `1.0.3` otherwise, you Claude may not update it correctly.

Yes, I have some other changes I'd like to make to streamline the test.  There's too much prompting at the moment.

PR #28, opened 2026-05-07
feat: add Gemini CLI support

### LINKED ISSUE #27: [Feature] Support for Gemini CLI
## Problem or use case

The Kobiton plugin currently supports **Claude Code** as MCP hosts. Users who prefer Google's [Gemini CLI](https://github.com/google-gemini/gemini-cli) have no first-class way to install or configure this plugin.

We want Gemini CLI users to be able to:

- Install the Kobiton plugin from the same repo without manual config gymnastics
- Authenticate to `https://api.kobiton.com/mcp` (OAuth 2.1 by default, or API key via `KOBITON_AUTH`)
- Invoke the same Kobiton tools (`listDevices`, `getSession`, `uploadAppToStore`, `run-automation-suite` skill, etc.) they already use in Claude Code / Copilot CLI

## Additional context

N/A

huytunguyenn:
## Summary

Add Gemini CLI extension support so the Kobiton plugin can be installed via `gemini extensions install` alongside the existing Claude Code and GitHub Copilot CLI install paths.

## Extension manifest

- Add `gemini-extension.json` at repo root with:
  - Inline `mcpServers.kobiton.httpUrl` pointing at the prod endpoint
  - `contextFileName: "AGENTS.md"` to reuse the existing instructions file (also read by Copilot CLI) — no separate `GEMINI.md` needed
  - Name `kobiton-automate` (Gemini convention: lowercase + dashes)
  - **No explicit `oauth` block** — Gemini CLI's default `authProviderType: "dynamic_discovery"` follows the same RFC 9728 `.well-known/oauth-protected-resource` flow that Kobiton MCP server already implements.

- Register `gemini-extension.json` in `scripts/validate.js` JSON-validity loop; update `scripts/validate.test.js` fixtures.

Skills (`skills/<name>/SKILL.md`) are auto-discovered by Gemini CLI from the default location — no manifest reference needed.

## Test plan

- [x] `pnpm run validate` passes (now also checks `gemini-extension.json`)
- [x] `pnpm test` passes (30 tests across 3 files)
- [x] Test: `gemini extensions link .` + `/mcp auth kobiton` + tool call end-to-end (after Gemini CLI installation)
- [x] Test: verify OAuth redirect URI Gemini uses is accepted by Kobiton's `OAUTH_ALLOWED_REDIRECT_URIS` allowlist (likely yes — both use loopback patterns)

## Notes for reviewers

- **Stacked PR** - base branch is `features/support-copliot-cli` (typo). GitHub auto-rebases to `main` when #10 merges. Until then, the diff here may show #10's frontmatter changes depending on how GitHub renders the comparison.

## Related issues

Closes #27

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,12 +7,31 @@
     "description": "Kobiton mobile testing platform plugins for AI coding assistants",
     "version": "1.0.0"
   },
-  "plugins": [{
-    "name": "automate",
-    "version": "1.0.2",
-    "description": "Connect to Kobiton mobile testing platform - manage devices, run automation suites, and view test results",
-    "category": "testing",
-    "tags": ["mobile", "testing", "appium", "automation", "devices", "kobiton"],
-    "source": "./"
-  }]
+  "plugins": [
+    {
+      "name": "automate",
+      "version": "1.0.2",
+      "description": "Connect to Kobiton mobile testing platform - manage devices, run automation suites, and view test results",
+      "author": {
+        "name": "Kobiton Inc."
+      },
+      "category": "testing",
+      "keywords": [
+        "mobile",
+        "test",
+        "automation",
+        "appium",
+        "mcp"
+      ],
+      "tags": [
+        "mobile",
+        "testing",
+        "appium",
+        "automation",
+        "devices",
+        "kobiton"
+      ],
+      "source": "./"
+    }
+  ]
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -8,6 +8,14 @@
   "license": "MIT",
   "repository": "https://github.com/kobiton/automate",
   "homepage": "https://kobiton.com",
-  "keywords": ["mobile", "testing", "appium", "automation", "devices", "kobiton"],
+  "keywords": [
+    "mobile",
+    "test",
+    "automation",
+    "appium",
+    "mcp",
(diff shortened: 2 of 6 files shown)


jeremylongshore:
Phase 1 reads cleanly. Particularly nice that dynamic-discovery OAuth lets the Gemini install flow piggyback off Kobiton's existing RFC 9728 `.well-known` endpoint — two CLIs now (with #10) installing against the same MCP server with zero server-side change.

Question on the manifest: skills auto-discover from `skills/<name>/SKILL.md` per the Gemini docs, but the spec also supports an explicit `contextFileName` array. Was on-demand agent invocation enough for the `run-automation-suite` use case, or did you weigh listing the skill files alongside `AGENTS.md` so they're loaded into initial context?

huytunguyenn:
@jeremylongshore Considered it briefly but went on-demand for several reasons:
1. Token cost: most sessions only need quick MCP tool calls, not the full Appium workflow.
2. Skill system designed for progressive disclosure: skill = on-demand workflow, should not make skill a part of the default system prompt.
3. `SKILL.md` is written in imperative invoke tone. Loading it upfront risks the model treating it as a directive: make it believe it's being asked to run automation from the first message.
4. `AGENTS.md` already references the skill by name, so discoverability is preserved without front-loading 1k+ tokens. 

If real usage shows the model is missing valid invocation opportunities, happy to revisit and gate the skill on `contextFileName` instead.

PR #29, opened 2026-05-11
feat(skill): default to device-only view URL

TranKienCuong:
## Summary

Update `run-automation-suite` Step 4 so the agent surfaces the **device-only view URL** (`<portal>/devices/launch?id=<deviceId>&view=device-only`) by default after launching a session. Falls back to the default-view URL only when the user explicitly asks to interact with the device (e.g. "let me drive it", "open the full session view").

`<deviceId>` continues to come from Step 2 (`listDevices` / `getDeviceStatus` / `reserveDevice`); no new MCP response fields are required.

The portal URL mapping is now derivation-based (`api*.kobiton.com` → `portal*.kobiton.com`) instead of a hard-coded per-environment table — naturally handles future environments without further skill updates.

## Why default to device-only

The device-only view is the cleaner default for agent-driven flows:

- The agent has already launched or referenced the session — the human is watching, not driving.
- Less surrounding UI = less visual noise for demos, screen-sharing, and embeds.
- If the user actually wants to interact with the device, they say so, and the agent emits the default-view URL.

## Version

`1.0.2 → 1.0.3`

## Test plan

- [x] Manual: ran `run-automation-suite`, confirmed the agent emits the device-only URL by default.
- [ ] Manual: re-run after merge with "open the full view" phrasing to confirm fallback path.


jeremylongshore:
Quick technical review — diff is small (+20/-6, 3 files) and the intent is sound:

**The change itself reads cleanly:**
1. Default to `?view=device-only` for agent-driven flows (cleaner watch experience, less surrounding UI).
2. Explicit fallback to default view on interaction-implying phrasing ("let me drive it", "open the full session view").
3. Portal URL is now derived (`api*.kobiton.com` → `portal*.kobiton.com`) instead of a hard-coded enum — forward-compatible across future environments without skill updates. Good call.

The trigger phrases in the fallback prose are clear enough for the model to disambiguate; this is a sensible default for the "agent has already launched, human is watching" path that `run-automation-suite` Step 4 actually serves.

**One coordination item with PR #52:** PR #52 (v1.2.0 release sync) also touches `.claude-plugin/plugin.json`, `CHANGELOG.md`, and `skills/run-automation-suite/SKILL.md`, and is open against the same base. If #52 lands first, this PR's 1.0.2→1.0.3 bump becomes a downgrade and you'll get three merge conflicts (one per shared file). Either:

- Merge this PR first (small, isolated), then rebase #52 — its CHANGELOG already lists v1.1.0 + v1.2.0, so adding a 1.0.3 entry between them is straightforward.
- Land #52 first, then rebase this onto `main` and bump to v1.2.1 (or fold the change into v1.2.0 directly if the timing works).

Either order is fine; just worth deciding before merging.

duchnguyen (approved):

Issue #55, opened 2026-05-18
[audit/R3] listSessions silently ignores limit parameter; server default page size returned regardless (F45)


jeremylongshore:
**Upstream mirror of [`jeremylongshore/automate#47`](https://github.com/jeremylongshore/automate/issues/47)** — content matches the fork issue. Filed here so Kobiton plugin team has the actionable server-side finding routed to the right owner.

---

## Finding F45

**Source:** External audit reply by @mimosa767 in [`kobiton/automate#53` (comment-4476188526)](https://github.com/kobiton/automate/issues/53#issuecomment-4476188526), filed 2026-05-18.

**Empirically confirmed 2026-05-18** via probe against `api.kobiton.com/mcp` (raw payload on file with Intent Solutions; available on request).

**Symptom:** `listSessions(limit=3)` returns 20 sessions. The `limit` request parameter is accepted without error but does not constrain the result count — the server returns its default page size (`current_size: 20`) regardless of the value passed.

## Repro

```
Tool: listSessions
Request: { "limit": 3, "userIntent": "..." }
Response (truncated):
  {
    "current_page": 1,
    "current_size": 20,
    "total_items": 100,
    "sessions": [ ... 20 items ... ]
  }
```

20 session objects returned instead of the requested 3.

## Severity for plugin DX

**High.** Three downstream impacts:

1. **Token budget.** Each session object expands to ~6 KB in some response shapes. Returning 20 instead of 3 sessions can push the response near the 25k-token MCP cap that R2's [upstream #35](https://github.com/kobiton/automate/issues/35) already documented as a truncation risk. A consumer asking for 3 sessions to stay well under the cap gets 20 sessions anyway.
2. **Programmatic correctness.** Any consumer using `limit` as a pagination primitive (slice first N, then call again with `offset=N` for the next page) gets nondeterministic behavior — every call returns 20, so pagination by `limit` produces overlapping result sets and never terminates as expected.
3. **Silent contract violation.** The parameter is documented and accepted; ignoring it without error masks the divergence from contract.

## Recommended fix

Honor the `limit` parameter in the `listSessions` server-side handler. Cap at a reasonable max (existing tools document max=20 for some endpoints; uniformity is fine) and reject values outside the range with a 400 + descriptive error rather than silent ignore.

Plugin-side mitigation is impossible — the parameter is forwarded to the server as-is and the server's response is what's truncated.

## Provenance

[...]

Issue #37, opened 2026-05-12
[audit/R2] Per-command session data + assertion semantics gap (F18 + F19 + F20)


jeremylongshore:
**R2 audit batch — bundle 4 of 5.** Part of the structured 5-experiment campaign run 2026-05-04.

## Scope

Three findings around session-data exposure and assertion semantics. **This bundle directly blocks #24 (`saveTestRun` + IQS test case CRUD).**

| Finding | What |
|---|---|
| **F18** | Per-command session data is not exposed in any of the 13 plugin MCP tools, despite the session record's `execution_data.all_command_data_available: true` flag advertising that the data exists. `getSession`, `getSessionArtifacts`, `listSessions`, `getApp` — none returns the per-step command stream. |
| **F19** | `getSession` advertises `command_screenshots_available: true` but no tool surfaces per-command screenshots. Same shape problem as F18. |
| **F20** | Assertion semantics are not exposed — pass/fail per command, assertion type, expected vs actual — even though they're part of the session record. Required for the `saveTestRun` use case (#24). |

## Severity

**High.** Blocks the saveTestRun + IQS test case CRUD feature (#24) — there's no path to "save this session as a reusable test case" without per-command data.

## Architectural fix

New tool: `getSessionCommands(sessionId, options?: { include: 'screenshots' | 'assertions' | 'all' })`. Returns the per-command stream advertised by `execution_data.all_command_data_available`. Keeps existing read-side tools unchanged (no breaking response shape).

## Canonical source

Full evidence + tool-shape spec + acceptance criteria lives in the fork issue: [`jeremylongshore/automate#23`](https://github.com/jeremylongshore/automate/issues/23).

Cross-reference: directly enables #24 (`saveTestRun` filed by @catjes).

jeremylongshore:
Closing this from Intent Solutions' side as the M1 engagement wraps.

This is a server-side feature ask (the `getSessionCommands` tool + the `saveTestRun` shape decision). Filed in R2 deliverable § 5.2 with the recommended ordered implementation: ship `getSessionCommands` first as an independently shippable plugin-DX win; then `saveTestRun` producing draft IQS test cases with blank assertions; then `annotateTestCase` as a separate flow; then the IQS test case CRUD endpoints. Plugin has no path around it.

Reopen as a server-side feature ticket if Kobiton wants to prioritize.

R2 deliverable: https://partners.intentsolutions.io/kobiton/r2/

jeremylongshore:
[...]

Issue #36, opened 2026-05-12
[audit/R2] W3C-strict /se/log silently breaks driver.getLogs + invisible deleteSession cooldown (F32 + F33)


jeremylongshore:
**R2 audit batch — bundle 3 of 5.** Part of the structured 5-experiment campaign run 2026-05-04.

## Scope

Two findings sharing the cross-tool-portability theme: silent log-endpoint failure + invisible cleanup state.

| Finding | What |
|---|---|
| **F32** | Kobiton's Appium endpoint is W3C-strict: `POST /se/log` (the legacy Selenium log-streaming path) returns "Unsupported URI". `driver.getLogs('logcat')` and `driver.getLogs('browser')` silently fail in any client that hasn't migrated to the W3C log API. Cross-tool portability is broken without a docs note or a session-time helper. |
| **F33** | After `deleteSession`, devices enter a ~5-minute cleanup cooldown. During cooldown, `reserveDevice` returns "device unavailable" (lumped via F22) with no signal that this is a transient post-cleanup state vs a permanent unavailability. |

## Severity

**High.** Both are silent-failure modes. Practitioners chase ghosts for hours before identifying the root cause.

## Architectural fix

- **F32**: docs note in `SKILL.md` capabilities reference, plus optional helper that swaps `/se/log` → `/session/{id}/se/log` for legacy callers.
- **F33**: `getDeviceStatus` returns `{ status, cooldown_until?, cooldown_reason? }` so the cooldown window is visible.

## Canonical source

Full evidence + W3C spec references + repro lives in the fork issue: [`jeremylongshore/automate#22`](https://github.com/jeremylongshore/automate/issues/22).

jeremylongshore:
Closing this from Intent Solutions' side as the M1 engagement wraps.

Plugin-side response shipped via fork PR #44 — the `session-error` advisory hook emits a span when W3C-strict /se/log is rejected; agents can route to alternate log paths. Server-side fix (either documenting the limitation in run-automation-suite/SKILL.md — cheap; or adding `cleanup_state` + `cleanup_until: ISO8601` to getDeviceStatus — richer) remains your call.

R2: https://partners.intentsolutions.io/kobiton/r2/ · R3: https://partners.intentsolutions.io/kobiton/r3/

jeremylongshore:
**Cross-link — backing upstream PRs:**

[...]

Issue #34, opened 2026-05-12
[audit/R2] confirmAppUpload async race + missing FAILURE_PARSING diagnostic body (F25 + F26)


jeremylongshore:
**R2 audit batch — bundle 1 of 5.** Part of the structured 5-experiment campaign run 2026-05-04 against `api.kobiton.com/mcp`. R2 anchor + reading order: see the R2 START HERE issue (will follow this one).

## Scope

Two findings sharing one architectural fix: extend `confirmAppUpload` response with state + diagnostic body.

| Finding | What |
|---|---|
| **F25** | `confirmAppUpload` returns 200 OK + a versionId BEFORE the platform's async parser has finished. The upload may still transition to `FAILURE_PARSING` afterward, and downstream tools (`createSession`, `reserveDevice`) fail with no clear root cause. |
| **F26** | When the upload does enter `FAILURE_PARSING` state, the response body is empty — no `parse_error` / `category` / `message` to tell the practitioner whether it was a manifest issue, a corrupted binary, a signature mismatch, or platform-side. |

## Severity

**High.** Combined effect: a failed upload looks like a successful upload for several minutes, then downstream tool calls fail without diagnostic detail.

## Architectural fix

`confirmAppUpload` returns `{ versionId, state, parse_error? }`. Practitioner / agent polls until state ∈ {`READY`, `FAILURE_PARSING`}. On `FAILURE_PARSING`, body carries `{ category, message }` so the cause is actionable in-band.

## Canonical source

Full evidence (raw `getApp` responses, timing data, repro steps, acceptance criteria, validation plan) lives in the fork issue: [`jeremylongshore/automate#20`](https://github.com/jeremylongshore/automate/issues/20).

R1 anchor for context: kobiton/automate#20 (now closed).

jeremylongshore:
Closing this from Intent Solutions' side as the M1 engagement wraps.

Plugin-side response shipped via fork PR #44 — the `post-upload` advisory hook emits an OTel span on the resolved upload state, which agents can route on until server-side notifications land. Server-side fix (extending `confirmAppUpload` with a `state` field + `parse_error: {category, message}` body for FAILURE_PARSING) remains your call.

R2 § 4 covers the architectural framing; R3 § 4 covers the advisory-hook design choice (multi-reviewer pre-flight cleared 6 BLOCKERs + 8 HIGHs by collapsing to advisory-only).

[...]

PR #67, opened 2026-05-19
feat(agents): add 3 specialized agents — capability reconciler, session triage, device picker

jeremylongshore:
## Summary

Adds three Claude Code subagents under `agents/` that decompose the most common in-context Kobiton operations into single-responsibility subagents. Each follows the Anthropic agent spec (`name` + `description` + `tools` allowlist) and is loaded automatically by Claude Code's `agents/` directory convention.

## What ships

- `agents/appium-capability-reconciler.md` — diffs parsed Appium capabilities against Kobiton's supported list; returns the minimal change set.
- `agents/device-picker.md` — parses a fuzzy device ask, calls `listDevices`, ranks, confirms before handing the UDID back to the parent skill.
- `agents/kobiton-session-triage.md` — fetches session artifacts and walks a detector ladder (parser race → cooldown collision → log-endpoint mismatch → capability mismatch → reservation conflict → script-level error) to narrow the failure.
- `skills/run-automation-suite/SKILL.md` § 2 + § Error Handling — adds a 2-retry exclude-and-rerank contract for `reserveDevice` `device_unavailable`.

## Addresses the reviews on this PR

**@duchnguyen** (2026-06-02 — agent `tools:` allowlist hygiene):
- Added `mcp__kobiton__listDevices` to `device-picker` allowlist (the only MCP tool the body invokes after the trim below)
- Added `mcp__kobiton__getSession` + `mcp__kobiton__getSessionArtifacts` to `kobiton-session-triage` allowlist (matches what Step 1 invokes)

**@TranKienCuong** (2026-06-04 — `device-picker` over-engineering):
- Step 3 scoring rubric → one-line procedure
- `listSessions` recent-sessions bonus → dropped; `mcp__kobiton__listSessions` removed from the allowlist
- Step 5 `reserveDevice` retry loop → moved to `SKILL.md § 2` with a 2-retry cap and the post-`terminateSession` cooldown context
- "Role & Expertise" preamble → dropped (the four-conflict-modes content stays as plain prose)
- "Sourcing discipline" section → trimmed to its one load-bearing sentence
- Separate "Error handling" → folded into the workflow steps
- `device-picker.md`: 93 → 58 lines

Also dropped issue URLs (`#33`, `#34`, `#36`, `#37`, `#55`) from all three agent bodies, `SKILL.md § 2`, and from `kobiton-session-triage`'s user-facing diagnosis output template. The behavior descriptions next to them already do the work in workflow prose. Same URLs still live in `skills/run-automation-suite/references/known-limitations.md` where the catalog is the right home for them.

## Checklist

- [x] `pnpm run validate` passes
- [x] `pnpm test` passes (53 tests)
- [x] Agent `tools:` allowlists match body invocations
- [x] DCO sign-off on commits

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/agents/device-picker.md
+++ b/agents/device-picker.md
@@ -8,7 +8,7 @@ tools: Read, Bash(node:*), mcp__kobiton__listDevices
 
 You translate a fuzzy natural-language device description into one specific Kobiton device UDID for the parent skill to reserve.
 
-`listDevices` filter surface: platform, platformVersion range, manufacturer, model, deviceName partial match, isOnline. Kobiton device-state semantics: `online`, `utilizing`, `reserved-by-self`, `reserved-by-other`. Per [`kobiton/automate#33`](https://github.com/kobiton/automate/issues/33), the four `device_unavailable` conflict modes are not always distinguishable upfront — choose conservatively when ranking.
+`listDevices` filter surface: platform, platformVersion range, manufacturer, model, deviceName partial match, isOnline. Kobiton device-state semantics: `online`, `utilizing`, `reserved-by-self`, `reserved-by-other`. The four `device_unavailable` conflict modes (offline / utilizing / reserved-by-other / pool exhausted) are not always distinguishable upfront — choose conservatively when ranking.
 
 ## When Claude Should Invoke You
 
@@ -30,7 +30,7 @@ Resolve the description into a filter triple:
 
 Call `listDevices` with the hard constraints.
 
-If the response is at or near the 25k-token cap (per [`kobiton/automate#55`](https://github.com/kobiton/automate/issues/55) — server pagination quirks), tighten the filter and re-query. Do not assume a truncated list is complete.
+If the response is at or near the 25k-token cap (server-side pagination may silently truncate), tighten the filter and re-query. Do not assume a truncated list is complete.
 
 If zero candidates: relax soft preferences and re-query. If still zero, hand back to the user.
 
@@ -51,7 +51,7 @@ Runner-up: Pixel 7a (UDID 9C432GGB1234) · Android 13 · online + available
 Reserve the top match? [y/n] or specify alternate
 ```
 
-On confirmation, return `deviceId` + `udid` + `platformName` + `platformVersion` to the parent skill. The parent owns the `reserveDevice` call and its retry contract (cooldown collision per [`kobiton/automate#36`](https://github.com/kobiton/automate/issues/36) is a likely cause of repeat `device_unavailable` failures).
+On confirmation, return `deviceId` + `udid` + `platformName` + `platformVersion` to the parent skill. The parent owns the `reserveDevice` call and its retry contract (the post-`terminateSession` ~5 min cleanup cooldown is a likely cause of repeat `device_unavailable` failures).
 
 ## Sourcing discipline
 
--- a/agents/device-picker.md
+++ b/agents/device-picker.md
@@ -1,93 +1,58 @@
 ---
 name: device-picker
-description: Translates a natural-language device request into a concrete Kobiton reservation. Queries listDevices, ranks candidates, confirms before reserving. Use when no UDID is given.
-tools: Read, Bash(node:*), mcp__kobiton__listDevices, mcp__kobiton__listSessions
+description: Translates a natural-language device request into a concrete Kobiton reservation candidate. Queries listDevices, ranks, confirms before handing off. Use when no UDID is given.
+tools: Read, Bash(node:*), mcp__kobiton__listDevices
 ---
 
 # Device Picker
 
-You are a specialized agent for Step 2 of the `run-automation-suite` skill: translating a fuzzy natural-language device description into a specific Kobiton device the test can run on.
+You translate a fuzzy natural-language device description into one specific Kobiton device UDID for the parent skill to reserve.
 
-## Role & Expertise
-
-You are expert at:
-
-- **The `listDevices` filter surface** — what each filter does (platform, platformVersion range, manufacturer, model, deviceName partial match, isOnline), and which combinations narrow effectively vs over-narrow.
-- **Kobiton device-state semantics** — what `online + utilizing + reserved-by-self + reserved-by-other` mean for picker logic. Per the documented platform behavior at [`kobiton/automate#33`](https://github.com/kobiton/automate/issues/33), the four conflict modes are not always distinguishable upfront; the picker must choose conservatively.
-- **Common test-target intent patterns** — "Pixel 7", "any modern Android", "latest iPhone with iOS 17", "a tablet for landscape testing", "the same device as last time" — each maps to a different filter shape.
+`listDevices` filter surface: platform, platformVersion range, manufacturer, model, deviceName partial match, isOnline. Kobiton device-state semantics: `online`, `utilizing`, `reserved-by-self`, `reserved-by-other`. Per [`kobiton/automate#33`](https://github.com/kobiton/automate/issues/33), the four `device_unavailable` conflict modes are not always distinguishable upfront — choose conservatively when ranking.
 
 ## When Claude Should Invoke You
 
-Invoke this agent when:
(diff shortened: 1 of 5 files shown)


jeremylongshore:
**Engagement-finding coverage:**

This PR ships three specialized agents that are plugin-side adjacent to the MCP server-primitives cluster — they encode parts of what would be `prompts/*` (F38) and orchestration-pattern advertisements (F36 `instructions`) on the agent-host side, since the server-side primitives are not yet declared.

- **Closed audit issues backed:** [`#39`](https://github.com/kobiton/automate/issues/39) (MCP server primitives cluster — F36 + F37 + F38 + F42).
- **F-findings adjacent:** F36 (`instructions` on `initialize`), F37 (`resources/*` capability), F38 (`prompts/*` capability). The agents are the host-side fallback while the server-side primitives are pending.

Full server-side recommendation for each finding: `docs/server-side-recommendations.md` § 5 shipping via [PR #77](https://github.com/kobiton/automate/pull/77).

- Jeremy Longshore
intentsolutions.io

agents/device-picker.md
```diff
@@ -0,0 +1,93 @@
+---
+name: device-picker
+description: Translates a natural-language device request into a concrete Kobiton reservation. Queries listDevices, ranks candidates, confirms before reserving. Use when no UDID is given.
+tools: Read, Bash(node:*)
```
duchnguyen: In Claude Code, the tools frontmatter in a subagent spec is an allowlist - tools not listed are blocked. The agent as written cannot call any of the MCP tools it depends on. Compare to how skills use allowed-tools with explicit MCP tool names.

Please add the required MCP tool names to the allowlist, e.g.:
tools: Read, Bash(node:*), mcp__kobiton__listDevices, mcp__kobiton__getDeviceStatus, mcp__kobiton__listSessions

agents/kobiton-session-triage.md
```diff
@@ -0,0 +1,107 @@
+---
+name: kobiton-session-triage
+description: Triages a failed Kobiton session — fetches artifacts and matches the failure against known R2 audit patterns (F25/F32/F33/etc). Use after non-success terminal state. Trigger with "triage session".
+tools: Read, Bash(curl:*), Bash(node:*)
```
duchnguyen: Please also add the required MCP tool names to the allowlist, e.g.: getSession(sessionId), getSessionArtifacts(sessionId)

agents/device-picker.md
jeremylongshore: Fixed in [`80d4409`](https://github.com/jeremylongshore/automate/commit/80d4409) — added `mcp__kobiton__listDevices` and `mcp__kobiton__listSessions` to the device-picker allowlist (the two MCP tools the agent body actually invokes — Step 2 calls `listDevices`, Step 3 ranking uses `listSessions`). Kept the allowlist minimal to avoid declaring tools the body never calls — let me know if you'd rather see `getDeviceStatus` included as an explicit add (your example listed it; the current body doesn't call it directly, so I left it out).

- Jeremy Longshore
intentsolutions.io

agents/kobiton-session-triage.md
jeremylongshore: Fixed in [`80d4409`](https://github.com/jeremylongshore/automate/commit/80d4409) — added `mcp__kobiton__getSession` and `mcp__kobiton__getSessionArtifacts` to the session-triage allowlist (Step 1 calls both). The agent's body also mentions `uploadAppToStore`, `getApp`, `terminateSession`, and `reserveDevice` but only as user-facing remediation recommendations, not as agent actions — left those out of the allowlist; happy to add if you'd rather the agent be able to invoke them directly.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
@duchnguyen one note re: the third agent in this PR — `agents/appium-capability-reconciler.md` has no MCP tool calls in its body (it operates on local files via `Read`, `Edit`, and `Bash(node:*) / Bash(python:*)`). Left its frontmatter unchanged. If that's not the right read, let me know.

- Jeremy Longshore
intentsolutions.io

TranKienCuong:
### Code Review

#### `device-picker` agent is over-engineered for what it does

**Confidence:** 75/100
**Location:** [`agents/device-picker.md`](https://github.com/kobiton/automate/blob/f48f8d97a4aab23345ddfe1dc581fab96606addc/agents/device-picker.md) (93 lines)

The agent's real job is small: parse a fuzzy device description → call `listDevices` → rank → confirm with user → return UDID. The spec inflates this into 93 lines with three layers that don't earn their keep:

1. **Fake-precision scoring rubric** (Step 3). The "availability ≫ match strength + recent-sessions bonus" rubric has no weights, no thresholds, no tiebreak rule — just vibes-ordering. An LLM reading this picks the obvious match without the rubric; the formalism is decoration.

2. **`listSessions` "recent successful sessions" bonus** (Step 3). This requires an extra MCP call to bias selection toward devices the user already used. But the agent's own "When NOT to Invoke" section excludes the case where the user pointed at a prior session — so the bonus fires on a path where the user *didn't* mention prior sessions. We're spending an API call to second-guess a fresh request.

3. **2-retry exclude-and-rerank reservation loop** (Step 5). This prescribes `reserveDevice` retry behavior, but the agent's `tools:` allowlist doesn't include `reserveDevice` — it can't execute the loop it specifies. That logic belongs in the parent skill that actually owns the reservation call.

**Suggested cuts:**

| Section | Action | Reason |
|---|---|---|
| "Role & Expertise" preamble | Drop | LLM doesn't need to be told it's an expert |
| Step 3 multi-factor scoring rubric | Replace with "rank by availability, then by match strength" | No weights = no rubric |
| Step 3 `listSessions` bonus | Drop | Extra API call for negligible lift |
| "Sourcing discipline" | Drop | Restates implicit "use live data, don't hallucinate" |
| Step 5 retry loop | Move to parent skill | Agent doesn't own `reserveDevice` |
| Separate "Error handling" section | Fold inline | Only the `device_unavailable` case mattered |

**What's worth keeping** (~30 lines instead of 93):
- 25k-token cap mitigation via filter tightening (real, MCP-specific, non-obvious)
- "Never auto-reserve, always confirm" (real policy)
- `device_unavailable` cooldown reference to issue #36 (real platform behavior)
- "Don't invent matches" guard (real LLM failure mode worth naming)

Compare to the other two agents in this PR: `appium-capability-reconciler` has real expertise (cross-language parsing + canonical three-policy rule) and `kobiton-session-triage` has a substantive detector ladder. `device-picker` is the thinnest of the three — the trim would put it on equal footing.

---

jeremylongshore:
Thanks for the review. Addressed each point in the latest push:

- Step 3 fake-precision rubric → one-line procedure
- `listSessions` bonus → dropped; `mcp__kobiton__listSessions` removed from the allowlist
- Step 5 `reserveDevice` retry loop → moved to `SKILL.md § 2` with a 2-retry cap + post-`terminateSession` cooldown context
- "Role & Expertise" preamble → dropped (kept the four-conflict-modes content as plain prose)
- "Sourcing discipline" → trimmed to its one load-bearing line
- Separate "Error handling" → folded into the workflow steps

`device-picker.md`: 93 → 58 lines. `pnpm run validate` + `pnpm test` clean.

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Closing this out cleanly.

The thesis behind these three agents — decomposing capability-reconciliation, session-triage, and device-selection into single-responsibility units — was the useful part, and I think it has been delivered. Since this PR opened, `automate` has consolidated on host-neutral skills (`drive-automation-session` and friends) rather than Claude-specific `agents/`, which is the right call for reaching Gemini / Copilot / Cursor / Codex from one source tree. Merging a top-level `agents/` directory tied to Claude Code's loader would cut against that direction, and this branch is several minor versions behind the current tree.

Thanks @TranKienCuong for the `device-picker` review — the over-engineering call was fair and I trimmed it, but the cleaner home for that behavior is a skill, not an agent. If `session-triage` in particular is worth having as a host-neutral skill rebased on the current release, I'm glad to bring it back in that shape — just say the word.

- Jeremy Longshore
intentsolutions.io

Issue #42, opened 2026-05-12
[audit/R3 above-spec] Server-side OpenTelemetry instrumentation — vendor-neutral observability across all MCP clients


jeremylongshore:
**R3 audit batch — bundle 4 of 4. Above-spec proposal, not a gap finding.**

## Why above-spec

The MCP 2025-06-18 spec doesn't mandate observability. But [issue #25](https://github.com/kobiton/automate/issues/25) by @frankwmoyer (regression test suite for agent experiences) and [PR #29](https://github.com/kobiton/automate/pull/29) by @TranKienCuong (liveViewUrl documentation) both signal that Kobiton wants to measure how AI agents use the plugin. The cleanest way to do that **across every MCP client uniformly** (Claude Code, Cursor, ChatGPT, Codex, Continue, Gemini CLI) is to instrument the **MCP server**, not each client.

## Scope

| What | Why |
|---|---|
| Wrap every \`tools/call\` / \`resources/read\` / \`prompts/get\` handler in an OTel span | Tool-call latency, error rate, retry pattern per agent type |
| Tag each span with semantic attributes | \`mcp.tool.name\`, \`mcp.client.name\`, \`mcp.session.id\`, \`mcp.user.org_id\`, plus tool-specific dimensions |
| Emit OTLP via \`OTEL_EXPORTER_OTLP_ENDPOINT\` env var | Standard OTel collector — works with Datadog, Honeycomb, Grafana Cloud, Tempo, Jaeger |

This is the vendor-neutral equivalent of Claude Code's \`claude_code.*\` spans. Claude Code can still emit its client-side spans on top; server-side spans become the ground truth visible to every client.

## Semantic attribute schema

\`\`\`text
# Span name: mcp.{method}  (e.g., mcp.tools.call, mcp.resources.read)

# Required attributes
mcp.protocol.version          string  e.g. "2025-06-18"
mcp.client.name               string  e.g. "claude-code", "cursor", "chatgpt-apps-sdk", "codex-cli"
mcp.client.version            string  client version
mcp.session.id                string  Mcp-Session-Id header value
mcp.tool.name                 string  e.g. "listSessions" (only on tools.call spans)
mcp.resource.uri              string  only on resources.* spans
mcp.prompt.name               string  only on prompts.* spans
mcp.user.org_id               string  Kobiton org/team
mcp.user.intent               string  userIntent field from request (110-145 chars)
mcp.tool.outcome              string  "success" | "tool_error" | "protocol_error"
[...]

Issue #40, opened 2026-05-12
[audit/R3] Resource subscriptions for async state — spec-layer fix for R2 F25+F26+F33


jeremylongshore:
**R3 audit batch — bundle 2 of 4.** Server-side architectural fix at the MCP spec layer for the async-state visibility R2 captured empirically as F25/F26/F33. Empirical precondition verified 2026-05-12: server currently declares no \`resources\` capability (see companion R3 bundle for the verified initialize response).

## Scope — re-fold R2 findings at the spec layer

| R2 finding (empirical) | Spec-layer fix |
|---|---|
| **F25** \`confirmAppUpload\` returns 200 OK + versionId BEFORE async parser finishes (#34) | Subscribe to \`kobiton://app/{appId}/state\`; server pushes state transitions via \`notifications/resources/updated\` |
| **F26** \`FAILURE_PARSING\` response body empty (#34) | Subscribed resource body carries \`{state, category, message_short}\` |
| **F33** \`deleteSession\` invisible ~5min cooldown (#36) | Subscribe to \`kobiton://device/{deviceId}/availability\`; server pushes \`cooldown_until\` and clears on cooldown expiry |

## Severity

**High.** Async state-change visibility is the difference between "agent waits and retries" (good) and "agent assumes failure and gives up" (bad). Works across every MCP client without Claude-specific hook code (Bundle 7 in our fork is the Claude-only fallback).

## Architectural fix (server-side)

Precondition: declare \`resources\` capability (companion bundle).

\`\`\`json
"capabilities": {
  "resources": {"subscribe": true, "listChanged": true}
}
\`\`\`

Expose subscribable resources via \`resources/templates/list\`:

\`\`\`json
{
  "resourceTemplates": [
    {"uriTemplate":"kobiton://app/{appId}/state","name":"app-upload-state","mimeType":"application/vnd.kobiton.app-state+json"},
    {"uriTemplate":"kobiton://device/{deviceId}/availability","name":"device-availability","mimeType":"application/vnd.kobiton.device-availability+json"},
    {"uriTemplate":"kobiton://session/{sessionId}/state","name":"session-state","mimeType":"application/vnd.kobiton.session-state+json"}
  ]
}
\`\`\`

Client flow:

\`\`\`json
// resources/subscribe (after confirmAppUpload)
{"method":"resources/subscribe","params":{"uri":"kobiton://app/v72107/state"}}

// Server emits when parser finishes
[...]

PR #70, opened 2026-05-19
feat(hooks): 3 advisory hooks for known platform timing windows

jeremylongshore:
## Summary

Adds three advisory Claude Code hooks for known platform timing windows around the plugin's MCP tool calls. All three are **advisory-only** — they emit text into agent context, never make HTTP requests, never read credentials.

## What's in this PR

| Hook | Event | Matcher | What it does |
|---|---|---|---|
| `advise-pre-terminate-cooldown` | PreToolUse | `terminateSession` | Allows the call but injects a notice about the device cleanup cooldown post-termination. |
| `advise-app-upload-poll` | PostToolUse | `confirmAppUpload` | Injects an advisory recommending the agent poll `getApp(appId)` until the async parser finishes. |
| `advise-post-terminate-cooldown` | PostToolUse | `terminateSession` | Confirms the cooldown window post-termination with `sessionId` for traceability. |

Plus `hooks/README.md` (operator guide) and `hooks/THREAT-MODEL.md` (T1-T9 across credential leakage, prompt injection, SSRF, PII echo, ReDoS, path manipulation, shell metachar handling, timeout DoS, supply chain integrity).

## Design constraint: advisory-only, no authenticated API calls

An earlier proposal had the hooks themselves call `getApp` to poll for state transitions. A multi-reviewer pre-flight (code-reviewer + security-auditor + test-automator) flagged it with multiple BLOCKERs — credential strategy, SSRF surface, PII echo from response body into agent context.

The current design eliminates the auth-needing surface entirely. **Hooks never make HTTP requests.** They emit text advisories into the agent's context window; the agent itself uses its already-authenticated MCP tools to do any real polling or follow-up. Same outcome, no security blast radius.

## Type of change

- [x] New feature
- [x] Documentation update

## Checklist

- [x] `pnpm run validate` passes
- [x] `pnpm test` passes (69 tests across 8 files; per-hook tests cover valid input, boundary cases, missing fields, malformed JSON, PII-leakage negatives)
- [x] CHANGELOG.md updated
- [x] DCO sign-off on commits

## Related

Cross-client observability (where applicable to Cursor / Codex / Gemini CLI / ChatGPT Apps SDK) is server-side OpenTelemetry, not client-side hooks — tracked separately at closed issue [#42](https://github.com/kobiton/automate/issues/42).

- Jeremy Longshore
intentsolutions.io


jeremylongshore:
**Engagement-finding coverage:**

This PR is the plugin-side deterministic-enforcement layer — four advisory PreToolUse/PostToolUse hooks plus the threat model. It is plugin-side adjacent to four async-state-visibility findings (server-side fixes documented in [PR #77](https://github.com/kobiton/automate/pull/77)) and provides Claude-Code-specific telemetry coverage of the above-spec OTel finding.

- **Closed audit issues backed:**
  - [`#34`](https://github.com/kobiton/automate/issues/34) (F25 + F26 — `confirmAppUpload` async race + `FAILURE_PARSING` body): `advise-app-upload-poll` PostToolUse hook
  - [`#36`](https://github.com/kobiton/automate/issues/36) (F32 + F33 — W3C `/se/log` + cleanup cooldown): `advise-pre-terminate-cooldown` + `advise-post-terminate-cooldown` hooks
  - [`#40`](https://github.com/kobiton/automate/issues/40) (resource subscriptions for async state — F25/F26/F33 fold at MCP spec layer): partial polling-style route-around
  - [`#42`](https://github.com/kobiton/automate/issues/42) (F44 — above-spec OTel): partial via `claude_code.hook` spans from the four handlers, narrow to Claude Code

- **F-findings covered:** F25 + F26 (mitigated via app-upload-poll), F32 + F33 (mitigated via terminate-cooldown advisories), F44 (partial Claude-Code-specific telemetry).
- **`validate-userintent` hook** is independent — enforces the documented `userIntent` format on every Kobiton tool call.

Full server-side recommendations for the underlying findings: `docs/server-side-recommendations.md` § 2 + § 7 shipping via [PR #77](https://github.com/kobiton/automate/pull/77).

- Jeremy Longshore
intentsolutions.io

hooks/scripts/validate-userintent.mjs
```diff
@@ -0,0 +1,96 @@
...
+    emitDecision('deny', GENERIC_REASON)
+    process.exit(0)
+  }
+
+  // Valid — exit 0 with no output. Claude Code interprets no output + exit 0
+  // as "no opinion, proceed normally".
+  process.exit(0)
+}
+
+main().catch(() => process.exit(0))
```
huytunguyenn: Thank you for the careful work on this. 

The scripts themselves are well-engineered (fail-open on bad input, no reflected user content in the deny reason, bounded regex). I'm requesting changes because of what `validate-userintent.mjs` *does*, not how it's code. 

**My main concern is that `validate-userintent` would deny almost every tool call for normal users, which is a blocker.**

This is a `PreToolUse` hook that emits `permissionDecision: "deny"` whenever `userIntent` doesn't match a fixed template:

```
[partner= exp-/]  | contact:   (110–145 chars)
```

That template is not a Kobiton requirement. On the plugin side, `userIntent` is a *required string* the agent generates from the user's prompt; the server doesn't ask for any partner/exp/contact format. 

Therefore, a normal call like:
```bash
> userIntent = `"List the user's available Android devices"`
```

doesn't match the regex (and is under 110 chars) -> **get denied**. 

In practice, an agent will essentially never produce the `[partner=... | contact:...]` shape on its own, so this hook blocks nearly every Kobiton tool call for anyone who isn't following this rule. Shipped to the public plugin, it makes the plugin unusable for the community.

A few related points on this file:
- `userIntent` is already a required field, so the presence check duplicates that. And the *format* gate adds a stricter contract the server never requested, which creates a client/server split (the hook denies inputs the server would accept) that will silently drift.
- Moreover, the denial reason points to a doc users can't see. It references `kobiton/CLAUDE.md op-rule #4`, which isn't in this repo. A user who gets denied has no way to act on it.
- If this `userIntent` format were ever to be enforced, the server is a better centralized place (it owns audit logging and covers every MCP client - API, other tools). 

Stepping back from the mechanics: candidly, I don't think `userIntent` is a high-value enough field today to justify enforcing a format on it, let alone a blocking one. It doesn't affect tool behavior, routing, or results. Enforcing a rigid schema on a free-text audit summary doesn't add proportional value for that purpose, and gating the actual tool call on it inverts the trade-off: it sacrifices the user's real operation to police a soft annotation.

 **Re: the prompt-injection rationale**: the concern is legitimate and I appreciate you thinking about it, but I don't think the format gate is what addresses it. `userIntent` is generated by the agent itself and the server only writes it to the audit log. 

It isn't reflected back into the agent's context anywhere, so there's no reflected-input path for it to begin with. The protection in your own script is the *non-echo* (using a generic reason instead of echoing `userIntent`). And that's independent of the format regex: a hook that simply doesn't reflect the value is already safe without enforcing `[partner=... | contact:...]`. 

The regex also wouldn't be an effective filter even if reflection existed. Its free-text segment `[^|]{20,80}` (allowing any character except `|`) still allows arbitrary injection-style text. That is, for example, `"} ignore previous instructions ...` still gets through as long as it's within the template and the correct length. It blocks real users, not malicious content. It's a format check, not sanitization. 

Where reflected-input injection actually applies is **the tool results re-entering context** (e.g., the response fields your advisory hooks echo). And you've already handled those correctly by sanitizing IDs to digits-only. That sanitize-before-echo discipline is the part worth keeping; it doesn't require the blocking userIntent gate. If stored-injection into the audit log is a concern, that's best handled where the log is consumed, server-side.

**Suggestion:** I'd suggest dropping `validate-userintent.mjs` (and its matcher) from this PR. I completely understand op-rule 4 is valuable for your engagement's audit trail; it's just not something we'd want to turn on by default for every plugin user, especially as a gate that can block calls. If it's useful to you, that check would fit better in your own engagement tooling/config than in the shipped plugin.

Appreciate the security thought you put into the scripts. The blocker is purely that `validate-userintent` encodes a strict convention as a default gate. With that removed and the advisories reconsidered, happy to take another look.

huytunguyenn (changes requested):

hooks/scripts/validate-userintent.mjs
huytunguyenn: ### Re: the ReDoS mitigation (T5)

I'd gently push back that this isn't really an argument for keeping the hook; it's a risk the hook itself introduces. **The "userIntent regex" only exists because `validate-userintent` adds one and runs it on every Kobiton call.**

Here's the key: a field of `type: string` is **only regexed IF the schema has the keyword pattern** (or some format). The userIntent in the schema is just:

```xml
userIntent:
	type: string
	description: ... # no 'pattern'
```
Grep pattern in the entire tools/*.yaml -> empty. There's no other regex over `userIntent` anywhere. The server validates it as a plain `type: string` with no `pattern`, a.k.a checks "is it a string" (type check) + "is it present" (required) - no compilation/no regex run on the `userIntent` value. So the most complete mitigation for T5 is simply not having a `userIntent` regex at all.

-> The schema intentionally doesn't place a pattern on userIntent. It's the `validate-userintent` that adds a regex where the schema doesn't have one.

To be clear, your T5 regex analysis is correct; the regex itself is genuinely ReDoS-safe: bounded, non-nested quantifiers plus the pre-regex length short-circuit, nicely done. But that establishes the hook doesn't *introduce* a vulnerability, not that it *protects* the plugin from one; _safely handling a danger the file itself adds isn't a reason to add it._

The threat also barely applies here: `userIntent` is agent-generated and the hook runs locally per call, so **a crafted value would spike the caller's own CPU, not a shared server**. Like the reflected-input point, I'd read this as a risk the hook creates and then handles carefully - solid engineering, but it argues for removing the hook rather than keeping it.

As a bonus, removing it also drops the per-call cost, which is a node process spawn + regex evaluation on every Kobiton tool call, since the matcher is `.*`.

So T5 ends up being one more reason `validate-userintent` is better removed than kept: dropping it eliminates both the (self-introduced) ReDoS surface and the per-call overhead.

jeremylongshore:
Good call — \`validate-userintent\` is engagement-internal discipline that doesn't belong as a default-on gate in the shipped plugin. Dropped the script, the \`PreToolUse\` matcher, and the README/threat-model references in the latest push; the format check moves into our own engagement-side tooling. Keeping the 3 advisory hooks (\`advise-pre-terminate-cooldown\`, \`advise-app-upload-poll\`, \`advise-post-terminate-cooldown\`).

- Jeremy Longshore
intentsolutions.io

huytunguyenn:
Hi @jeremylongshore, first, credit where it's due: **F25/F26 was a correct and valuable find.** The `confirmAppUpload` async race was real, and the new `getAppParsingStatus` tool in [v1.5.0](https://github.com/kobiton/automate/releases/tag/v1.5.0) exists as a direct result of you flagging it and keeping it on our radar (#34, this PR). 

Thank you, and the advisory-only redesign here was the right security call too. 🙌

On `advise-app-upload-poll` itself: the approach is sound. A client-side advisory is a reasonable way to mitigate this from the plugin side, and the hook hygiene in this PR (exec form, explicit timeouts, digit-only sanitization) is genuinely good. We've decided to fix it at the server instead, because that turns out to be the more thorough fix - and with it in place, **I'd like to drop this hook from the PR**. Two specifics:

**1. The guidance now ships in-band, in the tool response itself.** As of v1.5.0 (schema in #93, paired server deploy), `confirmAppUpload` returns a structured `next_step` block:

```jsonc
{
  "appId": null,        // null for a brand-new upload — see below
  "versionId": 423001,
  "state": "PARSING",
  "next_step": {
    "tool": "getAppParsingStatus",
    "args": { "versionId": 423001 },
    "poll_until_state_in": ["OK", "FAILURE_PARSING", "FAILURE_MISMATCHED_APP_IDENTITY", "FAILURE_SCANNING"],
    "description": "App parsing runs asynchronously. Poll getAppParsingStatus with versionId=423001 until ..."
  }
}
```

This reaches **every MCP host** (Copilot, Gemini, Codex, Cursor, ChatGPT - not just Claude Code, **the only host that runs these hooks**); it's machine-actionable rather than prose, and it can't drift from the server contract because it *is* the server contract. A `PostToolUse` advisory on top would inject a second, redundant copy of the same instruction into context.

**2. The `getApp(appId)` polling route couldn't work for the main case** - a detail that's genuinely hard to see from outside the backend: 
- `confirmAppUpload` returns `appId: null` for a brand-new app. The ID only exists once parsing creates the app record.
- It's non-null only when the upload is a new *version* of an existing app. Then we use `versionId` instead of `appId` to distinguish the new version of an existing app.

So the advisory's `appId=${appId}` interpolation renders empty for first uploads, and `getApp(null)` fails. Parse state has to be keyed by `versionId`, which is exactly what `getAppParsingStatus(versionId)` does. It also resolves the real `appId` once parsing lands. (Terminal states are `OK` / `FAILURE_*`; `READY` was never a state on this path.)

So: F25/F26 I'd consider closed by the server-side mechanism, with this PR as its origin story. If you rescope to the two terminate-cooldown advisories (`advise-pre-terminate-cooldown` + `advise-post-terminate-cooldown`), we'll review that on its own merits; the cooldown window is still a real timing gap on our side.

Thanks again for the find and the persistence here 🙏. This is exactly the kind of contribution that improves the plugin even when the final fix lands in a different layer 💪.

jeremylongshore:
Thanks @huytunguyenn — and congrats on v1.5.0. Agree completely on all three points.

The `next_step` block in the tool response is the right layer for this: it's machine-actionable, it can't drift from the server contract because it *is* the contract, and it reaches Copilot / Gemini / Codex / Cursor — not just the one host that runs hooks. A `PostToolUse` advisory echoing the same instruction would just be a redundant second copy.

And thanks for the `appId: null` detail — that's exactly the kind of thing invisible from outside the backend. It also retroactively explains a finding from the review cycle: `getApp` polling would've rendered `getApp(null)` on every first upload. Keying parse state on `versionId` is the correct fix, and `getAppParsingStatus` resolving the real `appId` on the way out is a nice touch.

Pushed a rescope: dropped `advise-app-upload-poll` (script + test + matcher, both mirrors), updated the README and threat model to the two remaining advisories, and left a short note pointing readers at v1.5.0 / `getAppParsingStatus` so the history is legible. What's left is the two terminate-cooldown advisories — the ~5min device cleanup window is still a client-side timing gap on our side, so they stand on their own merits.

Happy to call F25/F26 closed by the server-side mechanism. Appreciate the careful read on the security side too. 🙏

- Jeremy Longshore
intentsolutions.io

jeremylongshore:
Closing this out — and I think we already reached the conclusion together in the thread.

As @huytunguyenn laid out, the durable home for these timing-window advisories is the server-side `next_step` block (shipped in `getAppParsingStatus`, v1.5.0), not client-side hooks: it is machine-actionable, it cannot drift from the server contract because it *is* the contract, and it reaches every host rather than Claude Code only. That is strictly better than what these three advisory hooks were reaching for. `automate` also uses hooks only for CLI install (`SessionStart` → `install-cli.sh`), so a PreToolUse/PostToolUse advisory pattern would be a new convention the project has not adopted.

The finding landed where it counts — appreciate the explicit credit on F25/F26, and congrats again on v1.5.0. Closing rather than leaving a stale branch against a tree that has moved on.

- Jeremy Longshore
intentsolutions.io

PR #64, opened 2026-05-19
docs(skill): add Authentication section + Write/sleep to allowed-tools

jeremylongshore:
## Summary

Two narrow additions to `skills/run-automation-suite/SKILL.md`:

1. New `## Authentication` section (inserted after Prerequisites) documenting the three `.mcp.*.json` config variants the plugin ships with — so the agent has explicit auth-mode guidance at the top of the workflow rather than discovering it mid-run.
2. `allowed-tools` extended with `Write` and `Bash(sleep:*)` — both required by the existing skill body but missing from the allowlist.

Both changes are additive — no existing content rewritten.

## Changes

- `skills/run-automation-suite/SKILL.md` frontmatter: add `Write` and `Bash(sleep:*)` to the `allowed-tools` block-scalar
- `skills/run-automation-suite/SKILL.md` body: new `## Authentication` section between Prerequisites and Instructions, with a 3-row table of the config variants + the abort-on-no-MCP-connection guidance
- `CHANGELOG.md`: noted under `## Unreleased`

## Type of change

- [ ] New tool (added to `tools/*.yaml`)
- [ ] New skill (added to `skills/`)
- [ ] Bug fix
- [x] Documentation update
- [ ] Validation/test update
- [ ] Plugin metadata update

## Checklist

- [x] `pnpm run validate` passes
- [x] `pnpm test` passes (no test files changed)
- [ ] New tools have `title`, `annotations`, and `inputSchema` (n/a — no new tools)
- [x] Existing skill has `name` and `description` (unchanged); frontmatter additions are within the existing skill
- [x] CHANGELOG.md updated (user-facing change)
- [ ] README.md updated (n/a — README is unchanged; skill-internal documentation only)

## Related issues

Refs: #53

## Why the `allowed-tools` changes

Without `Write` and `Bash(sleep:*)` in the allowlist, the agent has to ask the user permission mid-workflow which breaks the documented no-interruption execution flow:

- **`Write`** is needed alongside `Edit` for the auto-edit capability reconciliation in **Step 3** (must-match fields are autocorrected; this is a content-write to the user's test script when reconciling rendered capabilities against parsed-script capabilities).
- **`Bash(sleep:*)`** is needed for **Step 4**'s "wait 2 seconds (to allow the session to initialize on Kobiton)" before opening the browser.

Both are existing instructions in the skill body — the allowlist just hadn't caught up.

## Why the `## Authentication` section

[...]

Issue #35, opened 2026-05-12
[audit/R2] Read-side shape consistency: listSessions overflow + field divergence + xium URL asymmetry (F14 + F16 + F17)


jeremylongshore:
**R2 audit batch — bundle 5 of 5.** Part of the structured 5-experiment campaign run 2026-05-04.

## Scope

Three findings around shape consistency on the read-side tools. Single architectural fix: adopt one canonical field shape across both endpoints + add explicit pagination.

| Finding | What |
|---|---|
| **F14** | `listSessions(limit=15)` overflows the 25k-token plugin output cap implicitly. The cap is a Claude Code default; `listSessions` doesn't surface the limit to the caller, so practitioners discover the truncation at runtime when fields go missing or sessions silently drop from the list. |
| **F16** | `getSession` and `getSessionArtifacts` use divergent field-naming for the same data (e.g., `device_name` vs `deviceName`, `start_time` vs `startTime`). Agents can't reuse parsing logic across the two endpoints. |
| **F17** | `xium-portal.kobiton.com/v1/sessions/{id}/...` URLs are asymmetric between `liveViewUrl` (full Portal) and `deviceOnlyViewUrl` (`?view=device-only` suffix). Per PR #29 from @TranKienCuong this asymmetry is being documented — overlap with R2 scope. |

## Severity

**Medium.** Not blocking, but compounds debugging friction across every session-handling agent.

## Architectural fix

- **F14**: explicit `pageSize`/`pageToken` pagination on `listSessions`; default `pageSize` calibrated below the 25k-token cap.
- **F16**: adopt one canonical field-naming convention (camelCase, matching the rest of the MCP surface) across `getSession` + `getSessionArtifacts`. Breaking change — bundle under a minor version bump.
- **F17**: documented; convergence path discussed in PR #29.

## Canonical source

Full evidence + repro + field-by-field comparison lives in the fork issue: [`jeremylongshore/automate#24`](https://github.com/jeremylongshore/automate/issues/24).

jeremylongshore:
Closing this from Intent Solutions' side as the M1 engagement wraps.

F14 mitigated plugin-side in R3: `listSessions` default `limit` lowered 20 → 10 (fork PR #45, mirrored upstream as PR #64). F16 (field-name divergence between getSession and getSessionArtifacts) + F17 (xium URL asymmetry) require server-side consistency work; plugin has no path around them.

Reopen at your discretion if you want to keep tracking the server-side consistency items.

[...]

PR #73, opened 2026-05-19
fix(tools): lower listSessions default limit 20 → 10 (F14 mitigation)

jeremylongshore:
## Summary

R2 audit finding **F14** mitigation: lower `listSessions` default `limit` from 20 → 10 to stay under the 25k-token Claude Code MCP response cap.

## Background

R2 audit (2026-05-04) empirically observed that `listSessions` responses with verbose `execution_data` per session can approach or exceed the 25k-token client cap at the previous default of 20. When the cap is hit, fields or entire sessions can be silently dropped at the client — the agent receives a partial result without an explicit error surfaced.

Lowering the default to 10 provides headroom against the cap with verbose `execution_data` fields. Agents can still request a higher count explicitly via `limit`; this change only affects the implicit default.

## Files

`tools/sessions.yaml` — single field change on `listSessions.inputSchema.properties.limit.default` (20 → 10) plus expanded description noting the cap rationale.

## Validation

- `pnpm run validate` passes
- DCO sign-off included on the commit

## Relationship to engagement deliverables

- **R2 audit finding F14**: `000-docs/021-AA-AACR-r2-mid-cycle-deliverable-report.md` § 10
- **Upstream audit issue**: closes-references `kobiton/automate#35` (closed 2026-05-19)
- **R3 deliverable correction**: the R3 deliverable (`000-docs/049-AA-AACR`) originally claimed this mitigation shipped via PR #64 — that was incorrect (#64 is the SKILL.md Authentication section). This PR is the actual F14 mitigation. The R3 § 7.1 deliverable mapping has been corrected to reflect this.

## Forward note

R3 § 4 Bundle 1 (server-side spec-conformance) flags `notifications/list_changed` + cursor-based pagination as the spec-aligned long-term fix — once the server adopts those primitives, this default-lowering becomes a stopgap. Filed as upstream issue `#40` (now closed); details consolidated in the forthcoming `docs/server-side-recommendations.md` (separate PR).

Refs `kobiton/automate#35` (closed).

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/tools/sessions.yaml
+++ b/tools/sessions.yaml
@@ -49,7 +49,7 @@ tools:
         limit:
           type: integer
           description: Maximum number of sessions to return
-          default: 20
+          default: 10
         offset:
           type: integer
           description: Offset for pagination


tools/sessions.yaml
```diff
@@ -48,8 +48,12 @@ tools:
           description: Filter by platform
         limit:
           type: integer
-          description: Maximum number of sessions to return
-          default: 20
+          description: >-
+            Maximum number of sessions to return. Default lowered from 20 to 10
```
huytunguyenn: The `Default lowered from 20 to 10` feels like a change in history.
It belongs somewhere else (in the commit message / CHANGELOG / this PR body discussion - where you've already covered it), not in the live description.
This will go stale the next time the default changes.

huytunguyenn (changes requested):
Thanks @jeremylongshore , the `default: 10` value itself is correct, so this is just about the description text. Requesting one change before merge.

The expanded description reads like a changelog rather than a tool spec. A param `description` is advertised to the agent via `tools/list` and read on every call, so it should describe **current behavior**, not the history or rationale of the change:

- The F14 / 25k-cap rationale is maintainer context, not something the agent needs to call the tool correctly. It should be better placed in the commit/PR.
- Same thing, the `Default lowered from 20 to 10` is change-history, so it should not be in the live description.
- Also, one detail from the MCP server side is that it is correct that the default is 10, but it clamps `limit` to a **maximum of 20** on its side. Therefore, it's also slightly misleading as written: it tells agents they can raise `limit`, but the server caps it at **20**. To go beyond that, the agent should page with `offset`, not raise `limit`.

Could you trim it to just the steady-state behavior the agent needs? I suggest something like:

  ```yaml
  limit:
    type: integer
    description: >-
      Maximum number of sessions to return. Defaults to 10; the server
      caps this at 20. To retrieve more, page with `offset`.
    default: 10
```

tools/sessions.yaml
jeremylongshore: Fixed in [`2d13534`](https://github.com/jeremylongshore/automate/commit/2d13534) — reverted the limit description to the original short form ("Maximum number of sessions to return") and kept the substantive change `default: 20 → 10`. The 25k-token rationale moved to the commit message where it belongs. Same principle you raised on #76 — descriptions describe what the tool does, not history. Thanks.

- Jeremy Longshore
intentsolutions.io

Issue #60, opened 2026-05-18
[audit/R3] uploadAppToStore response carries internal v1/v2 contradiction in confirm_upload.description vs confirm_upload.path (F50)


jeremylongshore:
**Upstream mirror of [`jeremylongshore/automate#52`](https://github.com/jeremylongshore/automate/issues/52)** — content matches the fork issue. Filed here so Kobiton plugin team has the actionable server-side finding routed to the right owner.

---

## Finding F50

**Source:** External audit reply by @mimosa767 in [`kobiton/automate#53` (comment-4476188526)](https://github.com/kobiton/automate/issues/53#issuecomment-4476188526), filed 2026-05-18.

**Verification posture:** schema-only. mimosa767's report captured the contradiction from a live upload session. We deliberately did not trigger a destructive re-probe — uploading a real APK/IPA creates a partner-system record + S3 file we'd then need to clean up, and mimosa767's report is concrete enough to file on.

**Symptom (from mimosa767):** the `uploadAppToStore` tool response contains an internal documentation inconsistency in the `confirm_upload` object:

- `confirm_upload.description` references `POST /v2/apps`
- `confirm_upload.path` references `/v1/apps`

Both fields describe the same confirm step in the same response payload.

## Why this matters

Whichever consumer reads the response gets two contradictory hints about which API version to call next:

- A human reviewing the tool output reads `description` and calls v2
- A code-generated client reads `path` and calls v1
- A downstream agent constructing the next request from the response shape may pick either, depending on which field it indexes

Both can't be right. The `confirmAppUpload` tool that the agent actually calls next (per `automate/tools/apps.yaml` lines 75-105 in this repo) doesn't take a version parameter — so this inconsistency is purely a downstream-confusion source, not a hard failure. But every downstream consumer is left guessing.

## Repro outline (for Kobiton internal verification)

1. Call `uploadAppToStore(filename, platform)` via the MCP tool (any small valid APK or IPA)
2. Inspect the full response payload, specifically the `confirm_upload` object
3. Confirm `confirm_upload.description` references a different API version than `confirm_upload.path`
4. Audit the rest of the response payload for additional v1/v2 references (in case the inconsistency is broader than the two fields named here)

## Recommended fix

[...]

Issue #59, opened 2026-05-18
[audit/R3] getApp.is_expired contradicts listApps.is_expired for the same app id (F49)


jeremylongshore:
**Upstream mirror of [`jeremylongshore/automate#51`](https://github.com/jeremylongshore/automate/issues/51)** — content matches the fork issue. Filed here so Kobiton plugin team has the actionable server-side finding routed to the right owner.

---

## Finding F49

**Source:** External audit reply by @mimosa767 in [`kobiton/automate#53` (comment-4476188526)](https://github.com/kobiton/automate/issues/53#issuecomment-4476188526), filed 2026-05-18.

**Empirically confirmed 2026-05-18** via paired probe against `api.kobiton.com/mcp` (raw payload on file with Intent Solutions; available on request).

**Symptom:** for the same app id, in the same minute, `listApps` and `getApp` disagree on `is_expired`. `listApps` reports the app as expired and carries the real expiry timestamp; `getApp` reports the app as not expired and drops the timestamp.

## Repro

```
App id: 678538 (name: "iOSPerf")
Probe date: 2026-05-18

listApps response (filtered to id 678538):
  { "id": 678538, "name": "iOSPerf", ...,
    "latest_version": {
      "id": 752776, "state": "OK",
      "created_at": "2026-02-25T08:03:34.695Z",
      "expiry_date": "2026-02-27T08:03:34.695Z",
      "is_expired": true
    }
  }

getApp(678538) response:
  { "id": 678538, "name": "iOSPerf", ...,
    "latest_version": {
      "id": 752776, "state": "OK",
      "expiry_date": null,
      "is_expired": false
    }
  }

Same app id (678538). Same version id (752776). Today (2026-05-18) is
nearly three months past the Feb 27 2026 expiry timestamp that
listApps reports. listApps' is_expired:true is the correct value;
getApp's is_expired:false is wrong.
```

Note: `getApp(686065)` (a FAILURE_PARSING app) returned a server 500 — confirmed separately as a likely error-path bug worth filing but out of scope for F49.

## Severity for plugin DX

**Critical.** Two endpoints describing the same underlying record disagree on a boolean that drives correctness. Consequences:

1. **Coin-flip behavior.** Any consumer that happens to read `getApp.latest_version.is_expired` to decide whether to upload a fresh build gets the wrong answer. Any consumer that happens to read `listApps[].latest_version.is_expired` gets the right answer. There is no observable signal in either response to tell the consumer which is authoritative.
[...]

Issue #58, opened 2026-05-18
[audit/R3] getDeviceStatus returns 3 fields; tool description documents battery, session info, and connection state which are absent (F48)


jeremylongshore:
**Upstream mirror of [`jeremylongshore/automate#50`](https://github.com/jeremylongshore/automate/issues/50)** — content matches the fork issue. Filed here so Kobiton plugin team has the actionable server-side finding routed to the right owner.

---

## Finding F48

**Source:** External audit reply by @mimosa767 in [`kobiton/automate#53` (comment-4476188526)](https://github.com/kobiton/automate/issues/53#issuecomment-4476188526), filed 2026-05-18.

**Empirically confirmed 2026-05-18** via probe against `api.kobiton.com/mcp` (raw payload on file with Intent Solutions; available on request).

**Symptom:** the `getDeviceStatus` tool description text says it returns "real-time status of a specific device including availability, current session info, battery level, and connection state." The actual response returns three fields and none of the latter three categories.

## Repro

```
Tool: getDeviceStatus
Request: { "deviceId": 1884315031, "userIntent": "..." }
Response: { "deviceId": 1884315031, "is_booked": false, "is_online": true }

Documented categories per tool description:
  - availability       → covered by is_booked + is_online ✓
  - current session info → no session_id / active_session / current_session field ✗
  - battery level      → no battery_level / battery_percent field ✗
  - connection state   → no connection_state / network_state / connection_type field ✗
```

## Severity for plugin DX

**High.** The tool exists specifically so the agent can decide whether to reserve a device. The current 3-field response answers "is this device free right now?" but not the follow-up questions that gate a useful reservation:

- **Battery level.** A device under 20% will die mid-session. A device above 80% will charge during the session and may show a low-battery overlay that breaks UI tests. Battery state is decision-critical pre-reservation; the current shape doesn't expose it.
- **Connection state.** Wi-Fi vs cellular vs no-network is the difference between a test scenario being valid and being unreproducible. Without this signal, the agent has to either probe inside a session (wastes a reservation) or trust the device-name string.
- **Current session info.** When `is_booked: true`, the agent has no way to know whose session is holding it or how long it's been held. Useful for self-coordination across CI workers, useful for debugging "why is this device perpetually unavailable."

Today's 3-field response delegates all of this discovery to a real reservation attempt, which is exactly the cost the tool is supposed to amortize.

## Recommended fix

Extend the `getDeviceStatus` response shape to include the three missing categories from the tool description:

```json
{
[...]

Issue #57, opened 2026-05-18
[audit/R3] getSessionArtifacts response is missing the documented screenshots category (F47)


jeremylongshore:
**Upstream mirror of [`jeremylongshore/automate#49`](https://github.com/jeremylongshore/automate/issues/49)** — content matches the fork issue. Filed here so Kobiton plugin team has the actionable server-side finding routed to the right owner.

---

## Finding F47

**Source:** External audit reply by @mimosa767 in [`kobiton/automate#53` (comment-4476188526)](https://github.com/kobiton/automate/issues/53#issuecomment-4476188526), filed 2026-05-18.

**Empirically confirmed 2026-05-18** via probe against `api.kobiton.com/mcp` (raw payload on file with Intent Solutions; available on request).

**Symptom:** the `getSessionArtifacts` tool description (`tools/sessions.yaml` in this repo, and the MCP tool description text returned by the server) documents four artifact categories: video recording, device logs, **screenshots**, and test reports. The actual response returns three of four — screenshots are not surfaced anywhere in the payload.

## Repro

```
Tool: getSessionArtifacts
Request: { "sessionId": 8641102, "userIntent": "..." }
Documented (per tool description): "video recording, device logs,
                                    screenshots, and test reports"
Response top-level keys: sessionId, video, testReport, logs

Missing: no top-level "screenshots" key; no per-screenshot URL list
inside any of the returned objects.
```

The `logs` object includes a `crash_logs: []` array and a `xium_preview_url` (live-view replay). Neither of these is a screenshot category.

## Severity for plugin DX

**Medium.** Two practical impacts:

1. **Documented capability gap.** Agents and human users read the tool description as authoritative. When they call `getSessionArtifacts` expecting to get screenshots and the response shape doesn't include them, the agent either fabricates an answer ("no screenshots were captured for this session") or has to escalate ("I don't see a screenshots field in the response"). Both are bad outcomes.
2. **Artifact-triage workflow blocker.** R2 finding F30 (artifact-triage observations) noted that screenshots are the primary artifact agents want to inspect for visual regression / element-not-found triage. Today's `getSessionArtifacts` shape doesn't deliver them; the workflow R2 prototyped against has no underlying API support.

## Recommended fix

Either:

[...]

PR #76, opened 2026-05-19
docs(tools): align 7 tool descriptions with empirical server response (F45-F50)

jeremylongshore:
## Summary

Post-R3 doc-drift audit (2026-05-18) surfaced 6 places where `tools/*.yaml` descriptions diverge from what the server actually returns. Audit was triggered by Stephen Penn's external-audit reply on [`kobiton/automate#53`](https://github.com/kobiton/automate/issues/53#issuecomment-4476188526); each finding was empirically probed against `api.kobiton.com/mcp` and filed as a closed audit issue (`#55-#60`).

This PR narrows the affected tool descriptions to match the empirical server response and ships a new `docs/post-r3-tool-doc-drift.md` catalog with one section per finding.

## Findings covered

| F# | Tool | Drift | Closed issue |
|---|---|---|---|
| F45 | `listSessions` | `limit` parameter is silently ignored — server returns default page size regardless of value passed | [`#55`](https://github.com/kobiton/automate/issues/55) |
| F46 | `getSession` | Response has no `has_video` field; `video_url` may be null with no indicator to disambiguate "no video" from "field missing" | [`#56`](https://github.com/kobiton/automate/issues/56) |
| F47 | `getSessionArtifacts` | Description advertises "screenshots"; response payload does not surface a `screenshots[]` category | [`#57`](https://github.com/kobiton/automate/issues/57) |
| F48 | `getDeviceStatus` | Returns `deviceId` + `is_booked` + `is_online`; description lists battery, current-session info, connection state (all absent) | [`#58`](https://github.com/kobiton/automate/issues/58) |
| F49 | `getApp` / `listApps` | For the same app id in the same minute, `getApp.is_expired = false` while `listApps.is_expired = true` and only `listApps` returns the timestamp | [`#59`](https://github.com/kobiton/automate/issues/59) |
| F50 | `uploadAppToStore` | Response `confirm_upload.description` references `/v2/apps`; `confirm_upload.path` references `/v1/apps` (internal contradiction) | [`#60`](https://github.com/kobiton/automate/issues/60) |

## What this PR does

**Two-part fix:**

### 1. Conservative description amendments (7 tools)

Each amendment adds a parenthetical pointer to the `docs/post-r3-tool-doc-drift.md` catalog entry for the longer narrative. The amendments narrow the advertised response shape to match what consumers will actually observe today — no over-promising what the server does not yet return.

| Tool | YAML file | Lines added |
|---|---|---|
| `listSessions` | `tools/sessions.yaml` | 3 (F45 note) |
| `getSession` | `tools/sessions.yaml` | 3 (F46 note) |
| `getSessionArtifacts` | `tools/sessions.yaml` | 1 (drop "screenshots" from response list, add F47 note) |
| `getDeviceStatus` | `tools/devices.yaml` | 3 (narrow to actual 3 fields, F48 note) |
| `listApps` | `tools/apps.yaml` | 3 (cross-reference to F49) |
| `getApp` | `tools/apps.yaml` | 4 (F49 note + canonical-expiry pointer to `listApps`) |
| `uploadAppToStore` | `tools/apps.yaml` | 4 (F50 note + canonical-path pointer) |

### 2. New `docs/post-r3-tool-doc-drift.md` catalog

One section per finding with: symptom, empirical repro, plugin-side mitigation in this PR, and the recommended server-side change (cross-referenced to the consolidated `docs/server-side-recommendations.md` PR).

## What this PR does NOT do

**Server-side response-shape changes** (the underlying fix for each finding — e.g., adding a `has_video` field, surfacing `screenshots[]`, computing `is_expired` consistently between endpoints) are deliberately out of scope for this PR. They are folded into the consolidated [`docs/server-side-recommendations.md`](https://github.com/kobiton/automate) (separate close-out PR — filed alongside this one).

This split keeps each PR atomic: this one is a defensive plugin-side artifact that needs no server change to ship; the server-side recommendations file is a documentation-only ask folded with the rest of the M1-M3 server-side findings.

## Validation

- `pnpm run validate` passes (4 tools in apps.yaml, 4 tools in sessions.yaml, 4 tools in devices.yaml, plugin.json, marketplace.json, .mcp.json, run-automation-suite SKILL.md)
- DCO sign-off included on the commit
- Pure-additive content + narrow text-substitution edits only — no behavioral changes

## Relationship to engagement deliverables

- **R3 deliverable § 6** (`000-docs/049-AA-AACR`): post-R3 external-audit mirror of F45-F50. This PR is the plugin-side close-out for that section.
- **Source-of-truth catalog**: `docs/post-r3-tool-doc-drift.md` shipped by this PR. Each finding traces back to its closed `kobiton/automate` audit issue + the originating external-audit comment on `#53`.

Refs `kobiton/automate#55, #56, #57, #58, #59, #60` (all closed).

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/tools/devices.yaml
+++ b/tools/devices.yaml
@@ -55,8 +55,8 @@ tools:
   - name: getDeviceStatus
     title: Get Device Status
     description: >
-      Get the real-time status of a specific device including availability,
-      current session info, battery level, and connection state.
+      Get the real-time availability status of a specific device (e.g.
+      `is_booked`, `is_online`) and `deviceId`.
     annotations:
       readOnlyHint: true
       destructiveHint: false
--- a/tools/sessions.yaml
+++ b/tools/sessions.yaml
@@ -62,6 +62,7 @@ tools:
     description: >
       Get detailed information about a specific test session including commands
       executed, device info, desired capabilities, metadata, and test results.
+      `video_url` may be null; there is no separate `has_video` indicator.
     annotations:
       readOnlyHint: true
       destructiveHint: false
@@ -85,7 +86,7 @@ tools:
     title: Get Session Artifacts
     description: >
       Get download URLs for session artifacts including video recording, device
-      logs, screenshots, and test reports.
+      logs, and test reports.
     annotations:
       readOnlyHint: true
       destructiveHint: false


tools/sessions.yaml
```diff
@@ -5,6 +5,9 @@ tools:
     description: >
       List test sessions with filters for status, device, platform, and date
       range. Returns session ID, status, device info, duration, and timestamps.
+      Note: the server may return its default page size if the requested
```
huytunguyenn: This is stale now, the server now honors `limit` (and clamped to a max of 20).

Adding this note would advertise a bug that does not exist, and it directly contradicts your own #73, which is premised on `limit` taking effect. Please drop this addition.

tools/sessions.yaml
```diff
@@ -61,6 +64,9 @@ tools:
     description: >
       Get detailed information about a specific test session including commands
       executed, device info, desired capabilities, metadata, and test results.
+      The response includes a `video_url` field which may be null; no separate
```
huytunguyenn: The substance is correct, `video_url` can be null and there's no `has_video`. Happy to keep it; just drop the `(F46 in docs/…)` pointer per the summary. 
E.g. "…and test results. `video_url` may be null; there is no separate `has_video` indicator."

tools/sessions.yaml
```diff
@@ -83,7 +89,8 @@ tools:
     title: Get Session Artifacts
     description: >
       Get download URLs for session artifacts including video recording, device
-      logs, screenshots, and test reports.
+      logs, and test reports. Screenshots are not currently surfaced in this
+      response (F47 in docs/post-r3-tool-doc-drift.md).
```
huytunguyenn: This is correct. I’d suggest simplifying it by removing the extra wording and the addendum/pointer. We probably don’t need to call out unsupported cases here.

A description should state what the tool returns, not enumerate what it doesn't  since that set is open-ended, and "not currently surfaced" would need updating if the response shape ever changes. Listing the supported behavior is enough.

```suggestion
       logs, and test reports.
```

tools/devices.yaml
```diff
@@ -54,8 +54,10 @@ tools:
   - name: getDeviceStatus
     title: Get Device Status
     description: >
-      Get the real-time status of a specific device including availability,
-      current session info, battery level, and connection state.
+      Get the real-time availability status of a specific device. Returns
+      `deviceId`, `is_booked`, and `is_online`. Battery level, current
+      session info, and connection state are not currently surfaced
+      (F48 in docs/post-r3-tool-doc-drift.md).
```
huytunguyenn: Careful here, on the server side, this tool passes the device status payload straight through rather than returning a fixed three-field shape, so `{deviceId, is_booked, is_online}` is what one probe happened to return, not a guaranteed contract. 

Asserting battery/session/connection are "not surfaced" can become wrong for a different device or state. Suggest describing it as availability-oriented without pinning the exact field set:

```suggestion
      Get the real-time availability status of a specific device (e.g. `is_booked`, `is_online`) and `deviceId`.
```

(Even better would be to narrow the server-side description so the contract and the docs stay in sync.) 

Drop the pointer either way.

huytunguyenn (changes requested):

tools/sessions.yaml
jeremylongshore: Reverted in [`c5bae15`](https://github.com/jeremylongshore/automate/commit/c5bae15) — dropped the limit-honor note entirely. You're right that it both advertises a non-bug and contradicts #73. Sorry for the cross-PR coherence miss.

- Jeremy Longshore
intentsolutions.io

tools/sessions.yaml
jeremylongshore: Used your exact suggested wording in [`c5bae15`](https://github.com/jeremylongshore/automate/commit/c5bae15): `video_url` may be null; there is no separate `has_video` indicator. Substance kept, F46 pointer dropped.

- Jeremy Longshore
intentsolutions.io

tools/sessions.yaml
jeremylongshore: Reverted in [`c5bae15`](https://github.com/jeremylongshore/automate/commit/c5bae15) to the original short description. Took the principle (describe what is returned, not what is missing) and applied it across the three apps.yaml descriptions proactively — `listApps`, `uploadAppToStore`, `getApp` all had similar "may disagree" / "may carry contradictory" claims with F-pointers. Reverted those too.

- Jeremy Longshore
intentsolutions.io

tools/devices.yaml
jeremylongshore: Used your exact suggested wording in [`c5bae15`](https://github.com/jeremylongshore/automate/commit/c5bae15). Agreed — pinning the 3-field shape as a contract was wrong since the tool is a server pass-through. `docs/post-r3-tool-doc-drift.md` stays as the engagement record of the original observation.

- Jeremy Longshore
intentsolutions.io

huytunguyenn:
Thanks for the revision, @jeremylongshore. This is much cleaner. The description edits now read as plain present-tense behavior, which is exactly right:

- `getSession`: "`video_url` may be null; there is no separate `has_video` indicator" is accurate and to the point
- `getSessionArtifacts`: dropping "screenshots" matches the actual response
- `getDeviceStatus`: "real-time availability status … (e.g. `is_booked`, `is_online`)" is good: the `e.g.` keeps it illustrative instead of pinning an exact field set, and the over-promised fields are gone
- Dropping the `listSessions` note and the `(F## in docs/…)` pointers was the right call

huytunguyenn:
@jeremylongshore I also noticed that the latest force-push no longer includes the `apps.yaml` edits. That’s fine, since F49 is now handled server-side.

For additional context on the `expiry_date` / `is_expired` finding in F49: thanks for catching this. From the user’s perspective, an uploaded app build in the Kobiton App Repo does not expire, so those fields were never meant to describe the app itself. Instead, they reflected the transient state of a download link, which is why `listApps` and `getApp` could report different values for the same app.

To remove that ambiguity at the source, we will drop `expiry_date` / `is_expired` from the MCP response on the server side, so no plugin-side note is needed. Thanks for your contribution.

docs/post-r3-tool-doc-drift.md
```diff
@@ -0,0 +1,179 @@
...
+---
+
+## How to read this catalog
+
+Each finding above maps 1:1 to a closed `kobiton/automate` audit issue (`#55-#60`). The plugin-side `tools/*.yaml` description edits in this PR are conservative — they make the descriptions match what the server actually returns today, so consumers calling these tools see accurate documentation without depending on a server-side change shipping first.
+
+The server-side response-shape changes are folded into [`docs/server-side-recommendations.md`](server-side-recommendations.md) (separate close-out PR), where they sit alongside the broader R2 + R3 server-side recommendations.
+
+- Jeremy Longshore
+intentsolutions.io
```
huytunguyenn: Regarding the `docs/post-r3-tool-doc-drift.md`, the description edits in this PR are the part I'd keep. The doc file is the one thing I'd suggest pulling out. Not about the content's value, just where it lives. A bit of reasoning:

- `docs/` here is user-facing usage documentation: right now it holds `examples.md`, which are natural-language prompt examples for end users, linked from the README and the skills. 
- The `post-r3-tool-doc-drift.md` drift catalog's audience is really the audit/engagement side (finding IDs, deliverable references, "raw payload archives on file"), which is a different reader than someone browsing the plugin's `docs/`.
- The findings already have a home: **GitHub issues.**:
	- F45, F50 are filed as #55, #60 under the `[audit/R3]` convention, with the START-HERE index issues tying them together. Issues are better suited to findings than a committed markdown file. 
    - They're **closeable**, **linkable**, and they don't go stale **inside** the shipped repo. We've actually seen that drift already in this review: the F45 entry (listSessions `limit` ignored) and the F49 entry (treat `listApps` as canonical for expiry) are both now out of date because those were addressed server-side.
    - An issue just closes; a catalog in the tree silently goes wrong.
- `server-side-recommendations.md` is server-team guidance, so it fits better as issues (or with the server) than as a file in the plugin repo, and the cross-links between the two docs are currently broken anyway since that file isn't in the PR.
- This doc also hardcodes point-in-time snapshots that will rot.

So my suggestion is to drop `post-r3-tool-doc-drift.md` from this PR, and keep the description edits. The audit narrative fits naturally in your engagement deliverables, and the per-finding tracking already lives in #55 and #60.

With the doc removed, the description changes are good to merge. Thanks again!

huytunguyenn (changes requested):

docs/post-r3-tool-doc-drift.md
jeremylongshore: Done in [`624bc9b`](https://github.com/jeremylongshore/automate/commit/624bc9b) — `docs/post-r3-tool-doc-drift.md` removed; kept the three YAML description edits (getDeviceStatus, getSession, getSessionArtifacts) per your earlier approval. The catalog content stays in the engagement-side deliverables; the per-finding tracking lives in #55 and #60 as you noted. Good point on the F45/F49 entries being already stale — that's exactly the kind of catalog-rot that justified pulling the doc.

- Jeremy Longshore
intentsolutions.io

huytunguyenn (approved):

Issue #78, opened 2026-05-21
[Feature] Remove browser menus, windows from phone display


catjes:
## Problem or use case

We want the video of the phone screen played during a manual claude session (agentic) to be only the phone - its important to put it in a tab without the headers, menus, shortcuts, url bar etc. 

We want the new version of the display be able to take input from the screen and pass that through to the running session as a command such as a touch / swipe etc.  This should also be relayed back to claude if we can.  

## Proposed solution

We want to see a new headless browser window that displays just the phone (think the way xcode does). 

<img width="416" height="228" alt="Image" src="https://github.com/user-attachments/assets/0e7e6221-e2fe-4264-bb8f-17b13c85903e" />

## Flow
1. Claude starts an interactive session as we do today
2. Claude sends commands as it does today
3. If there is a pause the user may send a command (touch / tap, swipe to start) that is executed in the app
4. The output and command is sent to Claude if possible

PR #86, opened 2026-05-27
feat(run-automation-suite): chromeless launcher for device-only session view

TranKienCuong:
## Summary

Adds a **chromeless launcher** to the `run-automation-suite` skill's session-open step. When the skill resolves the **device-only** view URL (`?view=device-only` — the default), it now opens Chrome in `--app` mode (no tab strip, no URL bar, no bookmarks bar) and resizes the window to a phone-shaped frame at runtime — macOS via `osascript`, Windows via PowerShell `SetWindowPos`, Linux launches `--app` only (no portable cross-WM resize hook).

If Chrome / Chromium is not installed on the host, the launcher exits silently and the skill falls through to the existing browser-preference open path (Chrome / Safari / Firefox / Default). When the user explicitly asks to drive the device manually (skill resolves the URL without `?view=device-only`), the chromeless launcher is skipped and the existing browser-preference path is used instead — preserves today's behavior on the "I want the full Kobiton UI" branch.

## Changes

- **New shim scripts** in `skills/run-automation-suite/scripts/`:
  - `chromeless-launcher.sh` — dispatcher: argv parse, URL validation, `$OSTYPE` routing
  - `chromeless-launcher-mac.sh` — `osascript` resize loop with 10s poll, URL-substring window match, per-window `try`/`on error` so a stray DevTools window doesn't abort the iteration
  - `chromeless-launcher-windows.ps1` — PowerShell + inline `Add-Type` `SetWindowPos`; `EnumWindows` matches on the spawned `chrome.exe` process tree; `ArgumentList` passed as an array to preserve quoting
  - `chromeless-launcher-linux.sh` — Chrome `--app` + `--window-size` hint; falls back to default browser via exit-2 sentinel if neither `google-chrome` nor `chromium` is on `PATH`
- **`skills/run-automation-suite/SKILL.md` Step 5** restructured: the launcher is invoked first on the device-only branch. The previous "Which browser should I open the session in?" prompt + `open` / `xdg-open` table are demoted to a fallback that fires when (a) Chrome is absent (exit-2 sentinel), or (b) the URL branch is the manual-interaction form.
- **`allowed-tools`** extended to include `Bash(bash:*)`, `Bash(pwsh:*)`, `Bash(osascript:*)`.
- **`chromeless-launcher.test.js`**: 32 vitest cases covering arg parsing, exit-code sentinels (`64` usage / `2` Chrome-absent fallback / `0` fail-open), URL metacharacter rejection (defense-in-depth against bash-quoting-breaking chars), and positive-path acceptance for real portal URLs with `&` between query params.
- **`scripts/sync-version.js`**: small drive-by fix — the CHANGELOG regex `^##\s+(\d+\.\d+\.\d+)\b/m` over-matched `1.3.0` against package.json's `1.3.0-dev.3`, tripping the CI gate on every dev-version cut. Now accepts SemVer 2.0 pre-release suffixes. Regression test added.
- **`.codex/skills/run-automation-suite/**`**: auto-mirrored by `sync-codex-artifacts.js`.
- **Version bump** `1.2.2` → `1.3.0-dev.3` via the new `sync-version.js` workflow (edit `package.json`, then `pnpm run build:version` propagates to all sibling manifests).

## Exit-code contract (called out for reviewers)

| Code | Meaning | Skill response |
|------|---------|----------------|
| `0` | Chrome launched (resize may have logged a warning) | Continue |
| `2` | Chrome / Chromium not installed | Fall through to the default-browser fallback (Chrome / Safari / Firefox / Default preference) |
| `64` | Usage error (missing `--url`, bad metacharacter, etc.) | Surface to the user — the launcher is buggy |

The launcher never throws. Window-resize failure (e.g., macOS Automation permission denied → osascript error `-1743`) is logged with a System Settings hint and the script still exits `0`. The automation session is never aborted by a resize failure.

## Source

- GitHub issue: kobiton/automate#78
- Bundled per-OS shim scripts; no Electron / Tauri dependency

## Test Plan

- [x] Unit tests pass — `pnpm test` **86/86** (54 pre-existing + 32 new)
- [x] `pnpm run validate` green — codex mirror in sync, all 5 plugin manifests in lockstep at `1.3.0-dev.3`, all tool YAML valid
- [x] Empirical macOS verification on macOS 26: Chrome `--app` window opens at the requested geometry; resize works when host process has Automation permission for Google Chrome (System Settings → Privacy & Security → **Automation** — NOT Accessibility); fail-open path triggers `osascript -1743` and exits `0` when the permission is revoked.
- [ ] Empirical Windows verification (PowerShell `SetWindowPos` against the spawned `chrome.exe` process tree)
- [ ] Empirical Linux verification (`--app` mode launch; `--window-size` hint honored by Mutter, ignored by tiling WMs — both behaviors acceptable per design)
- [ ] Manual happy-path test on each OS (macOS / Windows / Linux)
- [ ] Manual Chrome-absent fallback test (rename Chrome, verify the existing default-browser path still works)

## Acceptable trade-offs (documented in `chromeless-launcher-mac.sh` header)

- **First-run macOS prompt**: "X wants to control Google Chrome.app" — one-time, sticky per host process × target. Acceptable default-on friction; the prompt text itself documents what's being granted.
- **Linux no-auto-resize**: no portable cross-WM hook for foreign-window geometry. Linux still gets `--app` mode (which strips tab strip / URL bar / bookmarks), just without the runtime resize — still a visual win over today's default-browser tab.
- **Permission terminology**: macOS requires **Automation** (NOT Accessibility). Accessibility is the stricter system-wide grant gating keystroke / mouse injection; the launcher does only window-geometry operations (`set bounds of window`) which only needs the lesser Automation right. Empirically verified — toggling Accessibility off had no effect on the launcher; toggling Automation → Google Chrome off triggered the documented `-1743` fail-open path.

## Checklist

- [x] Code follows project conventions (vitest pattern matches `render-capabilities.test.js`; SKILL.md voice consistent; allowed-tools schema preserved)
- [x] No sensitive data exposed (URL validation rejects scheme injection and bash-quoting-breaking metacharacters)
- [x] Tests added for new functionality (32 new vitest cases — arg parsing, exit-code sentinels, URL metacharacter rejection, positive-path acceptance for real portal URLs)
- [x] Codex mirror regenerated and committed (`.codex/skills/run-automation-suite/**`)
- [x] Plugin version bumped and synchronized across all 5 manifests via `sync-version.js`

---


skills/run-automation-suite/scripts/chromeless-launcher-mac.sh
```diff
@@ -0,0 +1,100 @@
...
+    exit 64
+    ;;
+  http://*|https://*) ;;
+  *)
+    echo "chromeless-launcher-mac: --url must start with http:// or https://" >&2
+    exit 64
+    ;;
+esac
+
+CHROME_BIN="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
```
duchnguyen: **[MEDIUM]** Mac shim only checks one hardcoded Chrome path — misses Chromium, Canary, Brave

The macOS shim hardcodes `CHROME_BIN` to `/Applications/Google Chrome.app/...` and exits 2 (fallback to default browser) if that exact path is absent. A user running Chromium, Google Chrome Canary (`/Applications/Google Chrome Canary.app/...`), or a user-local Chrome install will silently fall through to the default-browser path, defeating the `--chromeless` opt-in even when a compatible browser is present. The Linux shim handles this correctly with a `command -v` loop over candidates.

**Suggestion:**
```bash
CHROME_BIN=""
for candidate in \
  "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  "/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary" \
  "/Applications/Chromium.app/Contents/MacOS/Chromium"; do
  if [ -x "$candidate" ]; then
    CHROME_BIN="$candidate"
    break
  fi
done
if [ -z "$CHROME_BIN" ]; then
  echo "chromeless launcher requires Google Chrome or Chromium — falling back to default browser" >&2
  exit 2
fi
```

skills/run-automation-suite/scripts/chromeless-launcher.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
```
duchnguyen, line 1: **[HIGH]** All scripts duplicated across .codex/ and skills/ with no shared source

Every script (`.sh` × 3, `.ps1`, `.test.js`) exists verbatim in both `.codex/skills/run-automation-suite/scripts/` and `skills/run-automation-suite/scripts/`. These are currently identical, but any bug fix applied to one tree will diverge from the other. The test file only runs against the `skills/` path, so bugs silently survive in `.codex/`. If the two directories must coexist, they should share a single source via symlinks or a build step; if one is the canonical location, the other should be removed.

**Suggestion:**
```bash
# Option A: make .codex copies symlinks pointing to skills/ canonical files
ln -sf ../../../../../../skills/run-automation-suite/scripts/chromeless-launcher.sh \
  .codex/skills/run-automation-suite/scripts/chromeless-launcher.sh
# repeat for each script

# Option B: delete the .codex/scripts/ subtree and update any .codex SKILL.md
# references to point to skills/run-automation-suite/scripts/
```

skills/run-automation-suite/scripts/chromeless-launcher.sh
```diff
@@ -0,0 +1,98 @@
...
+usage() {
+  cat >&2 <<EOF
+usage: chromeless-launcher.sh --url <URL> [--width N] [--height N]
+                              [--orientation portrait|landscape] [--x N] [--y N]
+EOF
+}
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --url)         URL="$2"; shift 2;;
```
duchnguyen, line 54: **[MEDIUM]** Numeric arguments --width/--height/--x/--y not validated in bash shims

The bash shims assign `--width`, `--height`, `--x`, `--y` directly to variables without checking that they are positive integers. A non-numeric value causes `set -e` to exit on later arithmetic (`$((X + WIDTH))`), but exits with a confusing shell error rather than the documented exit 64. A zero or negative dimension would pass through silently and produce an invalid `--window-size=0,0` or negative osascript bounds.

**Suggestion:**
```bash
is_positive_int() { case "$1" in ''|*[!0-9]*) return 1;; esac; [ "$1" -gt 0 ]; }

# In the argument-parsing loop, after assignment:
--width)  WIDTH="$2"; is_positive_int "$WIDTH" || { echo "--width must be a positive integer" >&2; exit 64; }; shift 2;;
--height) HEIGHT="$2"; is_positive_int "$HEIGHT" || { echo "--height must be a positive integer" >&2; exit 64; }; shift 2;;
```

skills/run-automation-suite/scripts/chromeless-launcher-mac.sh
TranKienCuong: Fixed in da0a9c6. The mac shim now walks a candidate list mirroring the Linux shim's pattern:

```bash
CHROME_BIN=""
for candidate in \
  "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  "/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary" \
  "/Applications/Chromium.app/Contents/MacOS/Chromium" \
  "/Applications/Brave Browser.app/Contents/MacOS/Brave Browser"; do
  if [ -x "$candidate" ]; then CHROME_BIN="$candidate"; break; fi
done
```

User-local installs (e.g. `~/Applications/`) still aren't picked up — same constraint as before, and I'd rather not start searching the home directory at runtime. Documented in the launcher header comment as the trade-off.

skills/run-automation-suite/scripts/chromeless-launcher.sh
TranKienCuong: The duplication is intentional and enforced by tooling — symlinks were the first instinct here too, but they can't ship.

**Why we can't symlink.** Codex CLI's plugin installer at [`openai/codex` `core-plugins/src/store.rs` `copy_dir_recursive`](https://github.com/openai/codex/issues/11314) silently skips symlinks. If `.codex/skills/run-automation-suite/scripts/chromeless-launcher.sh` were a symlink, Codex users would install a plugin where the launcher doesn't exist on disk. The `.codex/` subtree must be real files mirroring `skills/`. The README comment at the top of `scripts/sync-codex-artifacts.js` documents this.

**How drift is prevented.** `scripts/sync-codex-artifacts.js --check` runs as part of `pnpm run validate` (i.e. the CI gate). It compares every file in `skills/` against its `.codex/` counterpart byte-for-byte and fails the build on any divergence. So in practice the two trees can't drift — but you're right that the unit-test surface didn't previously cover this.

**Closed the test-coverage gap in da0a9c6.** Added 10 new vitest cases to `chromeless-launcher.test.js`:
- 5 × `.codex/ mirror has <file>` — assert each launcher script exists in the mirror
- 5 × `.codex/ mirror is byte-identical for <file>` — assert `readFileSync` of both copies matches

These run on every `pnpm test` and would fail loudly if the trees ever diverged.

**When this duplication goes away.** The day `openai/codex#11314` ships and is widely rolled out, we can replace the mirror with symlinks. The header comment at the top of `sync-codex-artifacts.js` already calls this out as a TODO tied to the upstream issue.

skills/run-automation-suite/scripts/chromeless-launcher.sh
TranKienCuong: Fixed in da0a9c6. Added `is_positive_int` + `require_positive_int` helpers in the three bash shims (dispatcher, mac, linux) and a parity check in the PowerShell shim:

```bash
is_positive_int() {
  case "$1" in ''|*[!0-9]*) return 1;; esac
  [ "$1" -gt 0 ]
}
require_positive_int() {
  is_positive_int "$2" || {
    echo "chromeless-launcher: $1 must be a positive integer (got: $2)" >&2
    exit 64
  }
}

# In the case loop:
--width)  WIDTH="$2";  require_positive_int --width "$WIDTH";  shift 2;;
--height) HEIGHT="$2"; require_positive_int --height "$HEIGHT"; shift 2;;
--x)      X="$2";      require_positive_int --x "$X";          shift 2;;
--y)      Y="$2";      require_positive_int --y "$Y";          shift 2;;
```

Non-numeric, zero, or negative inputs now exit `64` with the documented "must be a positive integer (got: X)" message instead of tripping `set -e` on later arithmetic.

The Windows shim was partially safe by virtue of PowerShell's `[int]` cast rejecting non-numerics at param-binding time, but I added a zero/negative check for parity.

9 new vitest cases cover the rejection paths (3 cases × 3 bash shims).

duchnguyen (approved):

PR #82, opened 2026-05-22
feat(tools): Test Case Management MCP tool schemas (14 tools)

TranKienCuong:
## Summary

Adds **14 MCP tool schemas** in a new `tools/test-management.yaml` file covering the full Kobiton Test Case Management (TCM) workflow.

| Category | Tools |
|----------|-------|
| Test cases (5) | `saveTestCase`, `listTestCases`, `getTestCase`, `updateTestCase`, `deleteTestCase` |
| Test runs (4) | `createTestRun`, `listTestRuns`, `getTestRun`, `terminateTestRun` |
| Test suites (5) | `listTestSuites`, `getTestSuite`, `createTestSuite`, `updateTestSuite`, `deleteTestSuite` |

Each schema declares `name`, `title`, `description`, `inputSchema` (with required `userIntent`), and `annotations` (`readOnlyHint` / `destructiveHint`). Domain: `TEST_MANAGEMENT`.

## Schema design notes

- `updateTestCase.modifications[].action` enum mirrors the backend's modification-type enum: `update_info` / `modify_steps` / `assign_to_test_suites` / `unassign_from_test_suites`. The `modify_steps` action carries nested `test_steps` with step-level actions (`remove` / `replace_text` / `submit_element`).
- List tools (`listTestCases`, `listTestRuns`, `listTestSuites`) accept an optional `teamId` query param so callers can scope results to a specific team.
- `saveTestCase` converts a finished manual exploring session into a reusable test case. The test case is named after the source session; rename / describe / re-tag via `updateTestCase` action `update_info` afterward.

## Diff stat

1 new file, +589 lines.

## Test plan

- [ ] `pnpm run build` produces `dist/tool-definitions.yaml` without errors
- [ ] CI schema validation passes
- [ ] After deployment, verify the MCP server picks up the 14 new tools via `tools/list`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #91, opened 2026-06-01
docs(contributing): align validation steps with validator's auto-discovery

jeremylongshore:
## What

Aligns CONTRIBUTING.md § "Adding New Tools or Skills to Validation" with what `scripts/validate.js` actually does.

The previous steps 1-2 told contributors:

```
1. Add the filename to the `toolFiles` array in `scripts/validate.js`
2. Add the skill directory name to the `skillDirs` array in `scripts/validate.js`
```

But there are no such arrays to maintain. `validate.js` uses `readdirSync` to discover tool YAMLs (line 125) and skill directories (line 164) at runtime. A contributor following the old instructions would search for array literals in the validator, find only `const`-declared `readdirSync` results, and either be confused or hand-edit the wrong thing.

Pure code-vs-doc reconciliation — no policy change. The kept steps (`setupValidProject` fixture + verify with validate/test) are the only contributor-side actions still required.

## Why filed separately

I noticed two adjacent docs items while preparing this fix, but they are content decisions Kobiton owns rather than bugs — flagging here for your call rather than including in this PR:

1. **README.md § Tools** says "13 MCP tools across 4 domains" and lists Devices + Sessions + Apps + Account. That count is internally correct, but the README doesn't include the 14 Test Case Management tools added in v1.2.2 / PR #82 (`tools/test-management.yaml`). If Test Case Management is intentionally not yet surfaced to README users, no change needed. If it should be documented there, happy to file a follow-up.

2. **AGENTS.md** lists `run-automation-suite` but not `run-interactive-test`. AGENTS.md was authored in PR #71 after `run-interactive-test` already existed (PR #45), so the omission looks intentional. If you want cross-host hosts (Gemini / Codex / Copilot) to know about `run-interactive-test` via AGENTS.md, happy to file a follow-up — with the macOS / Linux-x64 binary constraint clearly noted.

Both items are pre-existing; neither is in scope for this PR.

## Verification

- `pnpm run validate` clean
- `pnpm test` 5/5 files, 53/53 tests passing
- No code changes — docs only

- Jeremy Longshore
intentsolutions.io

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,12 +88,10 @@ Each skill file must have YAML frontmatter with `name` and `description`.
 
 ### Adding New Tools or Skills to Validation
 
-When you add a new tool YAML file or skill directory:
+`scripts/validate.js` auto-discovers tool YAMLs and skill directories via `readdirSync` — no manual array maintenance is required. When you add a new tool YAML file or skill directory:
 
-1. Add the filename to the `toolFiles` array in `scripts/validate.js`
-2. Add the skill directory name to the `skillDirs` array in `scripts/validate.js`
-3. Update the test fixtures in `scripts/validate.test.js` (`setupValidProject` function)
-4. Run `pnpm run validate && pnpm test` to confirm
+1. Update the test fixtures in `scripts/validate.test.js` (`setupValidProject` function) if your addition exercises a validation path not already covered
+2. Run `pnpm run validate && pnpm test` to confirm
 
 ### Validation
 


huytunguyenn (approved):

huytunguyenn:
Approving. Clean catch, thanks. Confirmed against `scripts/validate.js`: tool YAMLs and skill directories are both discovered via `readdirSync`, so there are no `toolFiles` / `skillDirs` arrays to maintain. The old steps 1 and 2 described a registration workflow that is outdated.

On the two items you flagged: both are real drift, and yes, follow-ups would be welcome:
  1. **README** "13 MCP tools across 4 domains": it's 27 across 5 once `tools/test-management.yaml` (14 tools) is counted
  2. **AGENTS.md**: adding `run-interactive-test` (with the macOS / Linux-x64 binary constraint noted, as you suggested) makes sense for the cross-host hosts

Thanks for keeping them scoped out of this PR.

Issue #81, opened 2026-05-22
[Feature] MCP skill: drive an automation session from natural language


TranKienCuong:
## Problem or use case

Today, users who want an automated check on a real device must either
pre-write an Appium script or drive the session manually. There is no
path that lets Claude take a natural-language intent (e.g. "log in,
open settings, toggle dark mode, log out") and produce a running
automation session on a device — plus a saved test case out the other
end — without a hand-authored script.

## Proposed solution

Add a skill to this repo that lets Claude drive an automation session
end-to-end from natural language:

- The skill orchestrates session lifecycle: start session, capture
  screen, perform interactions, end session — all via Appium against a
  reserved device.
- Claude reasons in a loop: observe the screen → decide the next
  action → emit an Appium command → observe the result → repeat — until
  the user's intent is satisfied or a failure mode is hit.
- The resulting session is a standard automation session, so existing
  test-case authoring picks it up unchanged.

Tool surface, prompt design, and the exact loop shape are implementation
details. The skill should expose interactions in a way that lets future
natural-language element-resolution backends plug in without changes to
the skill's contract.

## Acceptance criteria

- A user can describe an intent in natural language and the skill drives
  it on a reserved device end-to-end.
- The captured session is a valid automation session retrievable via the
  existing session tools.
- The skill handles common failure modes (element not found, timeout,
  unexpected screen) and surfaces them back to Claude for re-planning.

## Closure note for #51

This supersedes #51 ([Feature] Support MCP manual-session primitive),
which has been closed. Driving an automation session from natural
language reaches the same end goal (Claude produces a captured session
that becomes a test case) without the MCP layer having to mirror the
canvas-session transport.

## Parent

#24

PR #94, opened 2026-06-10
KOB-53296: add drive-automation-session skill (v1.6.0)

TranKienCuong:
## Summary

New `automate:drive-automation-session` skill that drives an already-reserved Kobiton device from a natural-language intent via direct WebDriver HTTP (automation session type). Complements `run-interactive-cli-session` (the renamed `run-interactive-test` — CLI session type); does not replace it. The returned session id is consumable by `mcp__plugin_automate_kobiton__saveTestCase`, closing the test-case-authoring flow gap from #81.

## Architecture

`scripts/appium.js` — Node `node:https`-only Appium HTTP client. **No new package dependencies.** Five subcommand shapes:

- **Generic mode** — `node appium.js --method <GET|POST|DELETE> --url <path-after-/wd/hub> [--req-body ...]`. The AI host emits raw Appium calls per `references/endpoint-reference.md`. Auto-wraps flat caps in the W3C envelope on `POST /session`.
- **`screen`** helper — captures **both** `iter-N.xml` and `iter-N.png` by default; emits `{hash, xmlBytes, pngBytes}`. `--xml-only` and `--png-only` flags for token-controlled modes.
- **`actions`** helper — builds W3C `/session/{id}/actions` body for touch / swipe / key sub-types.
- **`touch-perform`** helper — wraps a step array for legacy MJSONWP `/session/{id}/touch/perform`.
- **`control`** subcommand — signals end-of-cycle (DONE / BLOCKED) without an HTTP call.

`SKILL.md` describes a turn-based three-branch per-turn pattern: each turn the host picks exactly one of `screen` (observe), an Appium call (act), or `control` (end). A decision guide maps each previous-turn outcome to the next branch — e.g., a failed act with `no such element` calls for another `act` (with a corrected selector), not a fresh `screen`. Three reference docs in `references/`: `endpoint-reference.md` (allowlisted endpoints + selector-construction guide + web-session NATIVE_APP-first guidance), `loop-discipline.md` (per-turn pattern, stuck patterns, reading errors, PNG-by-default rationale), `capabilities.md`.

## Key design decisions

- **`kobiton:scriptlessCapture: true`** on session create — gates platform-side action recording for `saveTestCase` consumability (per KOB-41142).
- **`appium:newCommandTimeout: 1800`** — 30 min, survives human-in-the-loop pauses.
- **Host-driven blocker detection** — no programmatic N_REPEAT / N_UNCHANGED thresholds. Host watches the screen hash across turns and emits `control --blocked` when stuck. `MAX_ITERS=100` is the only programmatic safety net.
- **Single exit code (0 always)** — every failure path writes `iter-N.error.json`; host classifies recoverable vs fatal per the documented Appium error catalog.
- **Selectors built from observed XML + PNG, never invented** — explicit prose guidance in `endpoint-reference.md` with 4 XML-to-call examples. PNG-by-default catches native overlays (Chrome notifications welcome card, OS dialogs) the XML source doesn't see.
- **Web sessions: drive from NATIVE_APP context** — the universal approach (touch tap at viewport coords via the `actions` helper) covers both clicks and media activation, routes through the scriptless-capture allowlist, and sidesteps the three chromedriver pitfalls (in-webview click failures, missing user-activation, unstable context names). `execute/sync` is a brief fallback for web-only operations.
- **Cleanup contract** — Bash `trap` issues `DELETE /session/{id}` (idempotent on 404); AI host follows up with `mcp__plugin_automate_kobiton__terminateSession` as a belt-and-braces guarantee.

## Pilot-feedback follow-up (commits c8330db, 495254e — no version bump)

A first end-to-end pilot drove a YouTube intent on a Pixel 8a (32 iterations, success — Kobiton session 8701508 captured with video and consumable via `saveTestCase`). A second review (pre-session screenshot showed Chrome's "notifications" welcome dialog blocking the first turn) revealed XML-only observation misses native overlays entirely. Both reviews informed these patches:

1. **`screen` captures PNG by default.** Previously XML-only with `--include-screenshot` opt-in; now writes both `iter-N.xml` and `iter-N.png`. New `--xml-only` and `--png-only` flags for token-controlled modes. Mutually exclusive.
2. **`appium.js` auto-wraps caps in the W3C envelope on `POST /session`.** `render-capabilities.js` emits flat caps; the hub requires `{capabilities: {alwaysMatch: ...}}`. The pilot hit a 400 on its first attempt and worked around it manually. +3 vitest cases.
3. **`endpoint-reference.md` "Web sessions" section rewritten to lead with NATIVE_APP context.** Universal approach (touch tap at viewport coords) for both clicks and media activation. `execute/sync` retained as a brief fallback for web-only operations.
4. **SKILL.md Step 1 emphasizes credential-export-once.** The pilot re-parsed `~/.kobiton/.credentials` on every Bash call — anti-pattern callout added.
5. **SKILL.md Step 0: Device + app selection — ask before picking.** Inherited from `run-automation-suite`.
6. **SKILL.md Step 3b: Offer live view after session start.** Inherited from `run-automation-suite`. Surfaces `<portal>/session-explorer/<kobiton-session-id>?view=device-only` and opens via `open` / `xdg-open` on user confirm.
7. **`loop-discipline.md` adds "Why PNG is captured by default"** with the Chrome notifications welcome dialog as the canonical example.
8. **`allowed-tools` gains `Bash(open:*)` + `Bash(xdg-open:*)`** for the live-view flow.

## Other changes in this PR

- **Renamed** `skills/run-interactive-test/` → `skills/run-interactive-cli-session/` (and codex mirror). Naming pattern is now `<verb>-<session-type>-session`. `run-interactive-cli-session` keeps its CLI surface unchanged; the rename only signals which session type it targets.
- **`render-capabilities.js`** gains two optional flags: `--newCommandTimeout <seconds>` and `--scriptlessCapture`. Both default off (existing callers unaffected). `drive-automation-session` passes both.
- **`commands/setup.md`** doc tweak: lists `drive-automation-session` as a consumer of the credentials file written by `/automate:setup` (only `run-interactive-cli-session` depends on the additional CLI symlink).
- **Version 1.4.2 → 1.5.0** — propagated to all 5 manifests via `pnpm run build:version`. `.codex/` mirror synced via `pnpm run build:codex`. CHANGELOG top entry covers the surface area.

## Test plan

- [x] `pnpm run test` — **154/154 passing** (up from 116). New cases in `appium.test.js` (40) and `render-capabilities.test.js` (5). Deleted `screen-hash.test.js` (4) and `validate-action.test.js` (17) — their subjects were removed in the architectural refactor.
- [x] `pnpm run validate` — all checks pass; codex mirror in sync; version 1.5.0 across all 5 manifests; CHANGELOG top entry matches.
- [x] Manual end-to-end pilot on Pixel 8a (Chrome / YouTube intent / 32 iterations, success) — informed commits c8330db + 495254e.
- [ ] Manual TC-13 (iOS happy path) on XCUITest device
- [ ] Manual TC-12 (25-min pause survival) — verify session is still alive after 25-min mid-loop pause
- [ ] Re-pilot post-patch on a fresh device to verify the Chrome welcome dialog is now visible in iter-1.png and dismissed cleanly
- [ ] CI green


TranKienCuong:
### Code Review

Reviewed against CLAUDE.md, project history, spec, and the new code. Three findings ≥80 confidence.

---

**1. AGENTS.md not updated for the new skill (and has a stale path from the rename)** — confidence 90

CLAUDE.md `## Cross-tool surface`: *"When extending a skill's workflow or known-limitations list, mirror substantive changes into AGENTS.md so non-Claude hosts stay current."*

AGENTS.md has no entry for `drive-automation-session`. Gemini CLI, Copilot CLI, and Cursor (which read AGENTS.md instead of `skills/*/SKILL.md`) get zero awareness of the new skill.

Separately, the rename `run-interactive-test` → `run-interactive-cli-session` left two stale paths on line 42:

https://github.com/kobiton/automate/blob/4c45608fd14a20d9f651a4258811826161f749ec/AGENTS.md#L40-L44

Suggested fix: add a `## drive-automation-session` section mirroring run-automation-suite's depth, and update the two paths on line 42 to `skills/run-interactive-cli-session/`.

---

**2. CHANGELOG adds a contributor-managed `## X.Y.Z` section instead of `## Unreleased`** — confidence 88

CLAUDE.md `## Commits & PRs` (CHANGELOG bullet): *"Maintainers manage. They keep a single `## Unreleased` section between releases and rename it to `## X.Y.Z - YYYY-MM-DD` at release cut. Contributors should not add new version sections in PRs."*

https://github.com/kobiton/automate/blob/4c45608fd14a20d9f651a4258811826161f749ec/CHANGELOG.md#L1-L5

Historical pattern (commit `94c0b25` from the 1.4.0 release cut): dev.N entries accumulate on the feature branch then a maintainer collapses them to `## X.Y.Z` at release. The current branch shipping `## 1.6.0-dev.1 - 2026-06-12` verbatim would be the first time a dev-suffixed entry lands directly in main's CHANGELOG.

Suggested fix: rename the header to `## Unreleased`. The maintainer renames at release cut. (If keeping `1.6.0-dev.1` is the intent, get explicit maintainer sign-off — `sync-version.js`'s regex was already widened in 1.4.0 to handle this, but the merge-time convention is what's at stake.)

---

**3. Webview detection misses Appium drivers that emit an XML preamble** — confidence 82

https://github.com/kobiton/automate/blob/4c45608fd14a20d9f651a4258811826161f749ec/skills/drive-automation-session/scripts/appium.js#L286-L298

`isWebview = /^\s*<html[\s>]/i.test(srcXml)` only matches when `<html` is the very first non-whitespace token. Some Appium drivers (UiAutomator2 chromedriver in particular) prepend `<?xml version="1.0" encoding="UTF-8"?>` to webview `/source` responses. When that happens:

- `isWebview` is `false`
- `iter-N.xml` is written as the **raw** 559KB body (no strip)
- `iter-N.full.xml` is **not** written, so the escape hatch documented in `references/endpoint-reference.md` "Web sessions" is unavailable

The 558KB → 51KB optimization is silently bypassed and the host falls back to the `grep > probe.txt` dance the PR claims to eliminate.

Empirically: the actual pilot session (Chrome / YouTube / Pixel 8a) returned plain `<html ...>` with no preamble, so the strip *did* fire on that pilot. But the regex is not robust to other driver-version combinations.

Suggested fix: tolerate a leading XML declaration:
```js
const isWebview = /^\s*(?:<\?xml[^?]*\?>\s*)?(?:<!DOCTYPE[^>]*>\s*)?<html[\s>]/i.test(srcXml)
```
A vitest case feeding a preamble-prefixed body to `cmdScreen` would lock this in.

---

#### Below threshold (informational, not blockers)

- **76** — `stripBase64Images` regex `]*\bsrc=...>` stops at any literal `>` in a later attribute value (e.g. `onclick="x>0"`). Rare in real DOMs but possible.
- **76** — `response-shapes.md` rename status `R091` (vs `R099`/`R100` on its siblings) — explained by content edits in the same commit; benign.
- **75** — `cmdScreen`'s `srcXml` could be `undefined` after a `JSON.parse` failure if a test harness overrides `process.exit(0)`. Production behavior fine; possible vitest mocking gap.

---

### Agents Used
- claude-md-auditor: 2 high-confidence findings (90, 88)
- bug-detector: 1 high-confidence finding (82) + 2 below-threshold notes
- history-analyzer: 1 above-threshold echo (88; confirmed against `94c0b25`) + 1 below-threshold rename observation
- spec-validator: clean (5 minor content-drift notes, all <60)

---

README.md
```diff
@@ -386,9 +386,9 @@ The script is idempotent - safe to re-run.
 | Skill | Description |
 |-------|-------------|
 | **run-automation-suite** | Guided workflow for app upload, device selection, local Appium script execution (Node.js, Python, .NET, Java), and result collection. |
-| **run-interactive-test** | Guided workflow for interactive testing using natural language. WebDriver actions, device operations (adb shell, logs, screen), file management (push/pull), and more. |
+| **run-interactive-cli-session** | Guided workflow for interactive testing using natural language. WebDriver actions, device operations (adb shell, logs, screen), file management (push/pull), and more. |
```
huytunguyenn, line 389: Should add `drive-automation-session` skill at here too.

CHANGELOG.md
```diff
@@ -62,7 +125,7 @@
 ## 1.2.0 - 2026-05-18
 
 - Multi-CLI support: install on GitHub Copilot CLI, Gemini CLI, and Codex CLI in addition to Claude Code
-- New `run-interactive-test` skill — natural-language WebDriver/device/file commands powered by the bundled `kobiton` CLI wrapper (macOS Apple Silicon binary included)
+- New `run-interactive-cli-session` skill — natural-language WebDriver/device/file commands powered by the bundled `kobiton` CLI wrapper (macOS Apple Silicon binary included)
```
huytunguyenn: Should not change the old CHANGELOGs. 
The `run-interactive-test` name is the correct name for those old releases.

Suggest adding a note in the new changelog entry stating that `run-interactive-test` has been renamed to `run-interactive-cli-session`.

AGENTS.md
```diff
@@ -39,7 +39,28 @@ For exploratory testing or repro work (not running a pre-written script):
...
+
+For agent-driven flows the user describes in plain language ("open YouTube and play the first world cup video", "log in then enable Bluetooth, then go home"). Different from interactive driving above — this one auto-pilots from observation to action without a human in the loop on each step. The result is a saveable test case.
+
+1. **Ask before picking the device and the live view** (the skill blocks here): which device + which observation mode (foreground live view vs background run). For the device, the same `listDevices` / `reserveDevice` flow as the other skills applies.
+2. **Render capabilities** via `skills/run-automation-suite/scripts/render-capabilities.js` with `--newCommandTimeout 1800` (30 min — survives human-in-the-loop pauses) and `--scriptlessCapture` (so the resulting session is consumable by `saveTestCase`).
+3. **Resolve credentials** once into the process env via `skills/drive-automation-session/scripts/load-credentials.sh` — exports `KOBITON_USER` / `KOBITON_API_KEY` / `KOBITON_PORTAL` from MCP `getCredential` (preferred) or `~/.kobiton/.credentials` (fallback). Per-call invocations stay flag-free.
```
huytunguyenn: I don't see the `load-credentials.sh` scripts here.

Moreover, since `KOBITON_USER` / `KOBITON_API_KEY` / `KOBITON_PORTAL` are environment variables that could affect other skills and scripts at runtime, I suggest moving this script to the shared top-level scripts folder for centralized control, rather than keeping it inside a specific skill.

AGENTS.md
```diff
@@ -39,7 +39,28 @@ For exploratory testing or repro work (not running a pre-written script):
...
+5. **If the user chose foreground**, open the device-only live view URL via `skills/run-automation-suite/scripts/chromeless-launcher.sh` (Chrome `--app` window sized per device class) with the default-browser fallback table for Safari / Firefox / Default.
+6. **Per-turn loop** — three branches per turn:
+   - `screen`: capture `iter-N.xml` (stripped webview DOM or raw native source) AND `iter-N.png` by default. Compute a screen-state hash for stuck detection.
+   - `act`: emit a raw Appium HTTP call (`POST /session/{id}/element`, `POST /session/{id}/actions`, `POST /session/{id}/touch/perform`, ...). Selectors come from the observed XML — never invented.
+   - `control`: signal end-of-cycle with `--done` (goal reached) or `--blocked` (genuinely stuck).
+7. **Cleanup**: a Bash `trap` issues `DELETE /session/{id}` (idempotent — 404 = success). Follow up with `terminateSession` to guarantee the platform-side session ends.
+
+The session id is consumable by `getSession`, `getSessionArtifacts`, and `saveTestCase` exactly like a session created by `run-automation-suite`.
+
```
huytunguyenn, line 59: nit: session id -> session ID. Same with several places, really minor, so you can ignore this if this takes too much effort.

huytunguyenn: or just `sessionId`

AGENTS.md
```diff
@@ -39,7 +39,28 @@ For exploratory testing or repro work (not running a pre-written script):
...
+2. **Render capabilities** via `skills/run-automation-suite/scripts/render-capabilities.js` with `--newCommandTimeout 1800` (30 min — survives human-in-the-loop pauses) and `--scriptlessCapture` (so the resulting session is consumable by `saveTestCase`).
+3. **Resolve credentials** once into the process env via `skills/drive-automation-session/scripts/load-credentials.sh` — exports `KOBITON_USER` / `KOBITON_API_KEY` / `KOBITON_PORTAL` from MCP `getCredential` (preferred) or `~/.kobiton/.credentials` (fallback). Per-call invocations stay flag-free.
+4. **Create the automation Appium session** via the Node-only `skills/drive-automation-session/scripts/appium.js` (no package deps — uses `node:https` directly). Returns the session id.
+5. **If the user chose foreground**, open the device-only live view URL via `skills/run-automation-suite/scripts/chromeless-launcher.sh` (Chrome `--app` window sized per device class) with the default-browser fallback table for Safari / Firefox / Default.
```
huytunguyenn: The `Default` at the end of `Safari / Firefox / Default` is a bit confusing (with the "D" sounding like an app or something like that), especially since the table is already called the `default-browser fallback table`. 

Suggest spelling it out as `system default browser` to avoid ambiguity.

AGENTS.md
```diff
@@ -39,7 +39,28 @@ For exploratory testing or repro work (not running a pre-written script):
...
+   - `act`: emit a raw Appium HTTP call (`POST /session/{id}/element`, `POST /session/{id}/actions`, `POST /session/{id}/touch/perform`, ...). Selectors come from the observed XML — never invented.
+   - `control`: signal end-of-cycle with `--done` (goal reached) or `--blocked` (genuinely stuck).
+7. **Cleanup**: a Bash `trap` issues `DELETE /session/{id}` (idempotent — 404 = success). Follow up with `terminateSession` to guarantee the platform-side session ends.
+
+The session id is consumable by `getSession`, `getSessionArtifacts`, and `saveTestCase` exactly like a session created by `run-automation-suite`.
+
+Detailed step-by-step instructions live in `skills/drive-automation-session/SKILL.md`. The endpoint allowlist + selector-construction guide live in `skills/drive-automation-session/references/endpoint-reference.md`; per-turn loop discipline (stuck patterns, error catalog, artifact layout) lives in `skills/drive-automation-session/references/loop-discipline.md`.
+
+This skill complements — it does NOT replace — `run-interactive-cli-session`. Use it for AI-driven flows; use `run-interactive-cli-session` for human-driven exploration via the CLI session type.
```
huytunguyenn: This kind of note affects which skill the agent should pick, so it'd be more effective near the top of the section (or in the skill description) where the agent makes that routing decision, rather than at the bottom after all the instructions. 

By the time the agent reads it here, it may have already chosen its approach.

huytunguyenn (changes requested):

huytunguyenn (commented):
There is still old name `run-interactive-test` at `commands/automate/setup.toml`.

huytunguyenn (changes requested):
At `skills/run-interactive-cli-session/SKILL.md` still use the old name too:

```md
---
name: run-interactive-test
description: >-
```

skills/drive-automation-session/SKILL.md
```diff
@@ -0,0 +1,354 @@
...
+If the user's intent strongly implies a platform (e.g., "test the iOS App Store flow", "open Chrome on Android"), it's OK to auto-pick — but state which device you're picking and why in one line before reserving, so the user can redirect.
+
+### App
+
+For an `app` testing session:
+
+- If the user referenced a build (`@build.apk`, "the latest staging build", `kobiton-store:vXXXXX`), use it. Verify the store reference exists via `listApps` if uncertain.
+- If they didn't reference one, ask: *"Which app should the session install? (path to a local `.apk`/`.ipa`, an existing `kobiton-store:vXXXXX` reference, or none if the intent is browser-only)."*
+- Upload local files via `uploadAppToStore` → `confirmAppUpload`.
+
```
huytunguyenn: Recently, I have introduced `getAppParsingStatus` tool as a step to make sure the app is ready before continuing the session. 

The reason is that `confirmAppUpload` doesn't guarantee the app is ready, it only kicks off async parsing on the backend.

```suggestion
  - Upload local files via `uploadAppToStore` → `confirmAppUpload` -> poll `getAppParsingStatus` until the state is terminal (`OK` = ready; stop on `FAILURE_*`).```

File: skills/drive-automation-session/SKILL.md [hunk 8]
```diff
@@ -0,0 +1,354 @@
...
+# Branch B: act — execute an Appium call. Writes iter-N.request.json + either
+#   iter-N.response.json (success) or iter-N.error.json (any failure). The
+#   error file contains the raw Appium body or {error, message} for usage
+#   errors; the host reads it to decide what to do next turn.
+node "$SKILL_DIR/scripts/appium.js" <YOUR-ARGV> --session-dir "$SESSION_DIR"
+# Examples of <YOUR-ARGV>:
+#   --method POST --url /session/$SESSION_ID/element --req-body '{"using":"xpath","value":"//Button"}'
+#   actions --session-id $SESSION_ID --type swipe --from-x 540 --from-y 1800 --to-x 540 --to-y 600
+#   touch-perform --session-id $SESSION_ID --steps @/tmp/steps.json
+#   --method DELETE --url /session/$SESSION_ID/element/$EL_ID/clear
```
huytunguyenn: nit: `/clear` endpoint should be `POST`, not `DELETE`.

huytunguyenn (changes requested):

skills/drive-automation-session/SKILL.md
```diff
@@ -0,0 +1,354 @@
...
+If the `DELETE` somehow fails silently, the session will time out on the Kobiton side per `appium:newCommandTimeout` (30 min, set in Step 2). Better to wait for that timeout than to mark the session `TERMINATED`.
+
+### 6. Return
+
+Emit the session id back to the caller. The session is now consumable by:
+
+- `mcp__plugin_automate_kobiton__getSession({sessionId})` — session metadata, device info, results.
+- `mcp__plugin_automate_kobiton__getSessionArtifacts({sessionId})` — video, logs, screenshots, reports.
+- `mcp__plugin_automate_kobiton__saveTestCase({sessionId})` — persist as a re-runnable test case (sibling feature, out of scope here).
+
```
huytunguyenn, line 335: I see a lot of places use this kind of prefix form `mcp__plugin_automate_kobiton__*`.

This is **valid for Claude Code's mechanism only**. Other platforms like Copilot, Codex, Gemini, etc. might use different naming entirely. Models can use fuzzy-match, so the prefix just costs tokens and potentially causes a tool-not-found retry.

Moreover, even with Claude Code, this naming is used when installed as a plugin. With the repo's own `.mcp.json` (or added via `claude mcp add`) the same tool registers as `mcp__kobiton__getSession`.

Suggest using bare tool names everywhere, with a one-time `the Kobiton MCP tool ...` qualifier per doc.

skills/drive-automation-session/references/endpoint-reference.md
```diff
@@ -0,0 +1,330 @@
...
+
+1. **`accessibility id`** — `content-desc` on Android, `name` on iOS (the accessibility label). Fastest lookup; intentionally exposed for testing; most stable across releases. Use whenever a `content-desc` (Android) or `name` (iOS) attribute is present and meaningful.
+
+2. **`id`** (resource-id on Android, `name` on iOS) — second-fastest. Stable across screens within the same app version, less so across versions. Note: this is the Appium `id` strategy, **not** the opaque element-id returned by `POST /element` (that one is for use in interaction calls, see below).
+
+3. **`xpath`** (relative > absolute) — slow but expressive. Prefer **relative xpath** (`//Button[@text='OK']`, `//*[@content-desc='settings_btn']`) over absolute paths (`/hierarchy/android.widget.FrameLayout[1]/.../android.widget.Button[3]`). Absolute paths are brittle — any structural change to the screen breaks them. Use relative xpath when the target has no `content-desc`/`id` but has a distinctive attribute (`text`, `name`, `value`, `label`, or a unique combination).
```
huytunguyenn: This example only mentions the Android button (while **class name** section below mentions both `android.widget.Button` and `XCUIElementTypeButton`).

skills/drive-automation-session/references/endpoint-reference.md
```diff
@@ -0,0 +1,330 @@
...
+
+`appium.js` reads `~/.kobiton/.credentials` (written by `/automate:setup`) on each invocation. No credential flags or env vars.
+
+## Building Appium calls from the observed XML
+
+Every Appium call that targets an element follows the same workflow: **observe → read the XML → build a selector from the XML → find element → use the returned element id in the interaction call**. Selectors are **not invented**; they come from attributes you can see in `iter-K.xml`. Inventing a selector (guessing an id or class name that "should" exist) is the most common cause of `no such element` errors.
+
+### Selector strategies, in preference order
+
+When more than one strategy works for a target, prefer the higher-listed one. The reasons are speed, stability across screens, and survivability when the app's UI changes.
```
huytunguyenn: This flat list mixes native and web strategies, which ends up ranking `css selector` below `xpath`.

For web contexts, from standard WebDriver guidance:

> On ChromeDriver, CSS selectors are generally faster and less brittle than XPath

Also, the `css selector` doesn't apply to native screens at all, which is why you have put "web context only" in bullet 4. Therefore, I suggest splitting by context: native vs web.

Suggest something like:
- Native context (UiAutomator2 / XCUITest):
	- accessibility id: ...
	- id: ...
	- xpath: ...
	- class name: ... 	
- Web context (browser session or `WEBVIEW_*`)
	- css selector: ...
	- xpath: ...

huytunguyenn (changes requested):

TranKienCuong:
Thanks for the thorough review @huytunguyenn — all addressed in `d434e69`. Summary:

**Rename completeness**
- `commands/automate/setup.toml` (both lines), `CLAUDE.md` (skills table + AGENTS note + setup symlink note), `README.md` footnote → all `run-interactive-test` → `run-interactive-cli-session`. The frontmatter you flagged in `skills/run-interactive-cli-session/SKILL.md` was already correct on the latest commit.

**CHANGELOG**
- Restored the historical `1.2.0` / `1.2.1` entries — you're right, `run-interactive-test` was the correct name at those releases; reverted my edits to them.
- Added a "Renamed: run-interactive-test → run-interactive-cli-session" note in the `1.6.0` entry instead.
- Also scoped the `1.6.0` entry down to what the release delivers (dropped the internal iteration narrative + the stale `--include-screenshot` / `--portal/--user/--api-key` mentions).

**README** — added `drive-automation-session` to the Skills table.

**SKILL.md**
- `confirmAppUpload` → now polls `getAppParsingStatus` until terminal before continuing. Good catch.
- `/clear` example fixed to `POST` (the endpoint table already had it right; only the example was wrong).

**MCP tool prefix** — dropped `mcp__plugin_automate_kobiton__` from all prose across SKILL.md + references + AGENTS.md; added a one-time "Tool naming" qualifier explaining the host resolves the prefix and models fuzzy-match the bare name. Agreed on the reasoning.

**AGENTS.md**
- The `load-credentials.sh` reference was stale — that script was deleted in an earlier security fix; credential loading now lives **inside `appium.js`** (reads `~/.kobiton/.credentials` directly, never via argv/env/transcript). So the "move to shared top-level scripts/ for centralized control" concern is resolved by there being no separate script — the logic is encapsulated in the one place that uses it. Let me know if you'd still prefer a shared helper module.
- "Default" → "system default browser"; "session id" → "session ID"; routing note moved to the top of the section; cleanup uses `DELETE /wd/hub/session` only (COMPLETE, not TERMINATED).

**endpoint-reference.md**
- Split selector-strategy preference into native vs web contexts; web now ranks `css selector` above `xpath` per the ChromeDriver guidance you cited.
- Added an iOS (`XCUIElementTypeButton name=...`) variant to the accessibility-id example.

**CodeQL** ("outbound network request depends on file data", `appium.js`) — this is a by-design false positive: the "file data" is the user's own `~/.kobiton/.credentials` (written by `/automate:setup`), and the hub URL is meant to be configurable per portal. Could you dismiss it as "won't fix — by design" from the Security tab? I don't have access to that UI.

176/176 vitest; validate + codex mirror clean.

.codex/skills/drive-automation-session/scripts/appium.js
```diff
@@ -0,0 +1,427 @@
...
+let currentFlags = null
+
+function fail(_unusedCode, error, message) {
+  process.stderr.write(JSON.stringify({error, message}) + '\n')
+  const base = currentFlags ? artifactBase(currentFlags) : null
+  if (base) persistError(base, 0, JSON.stringify({error, message}))
+  process.exit(0)
+}
+
+const softFail = (_base, error, message) => fail(0, error, message)
```
duchnguyen: **[MEDIUM]** softFail ignores its _base argument — latent defect if called early

`softFail` accepts `_base` as its first parameter but unconditionally delegates to `fail()`, which re-derives the artifact path from the global `currentFlags`. Call sites like `softFail(base, 'parse', ...)` (lines 255, 269) pass an already-resolved `base`, implying the caller believes that value will be used. Because `currentFlags` is set before any helper runs this is harmless today, but if `softFail` were ever called before `parseArgs` completes — e.g., during early validation refactors — the error artifact would be silently dropped with no indication that the passed `base` was ignored.

.codex/skills/drive-automation-session/scripts/appium.js
```diff
@@ -0,0 +1,427 @@
...
+    const req = (isHttps ? httpsRequest : httpRequest)({
+      protocol: u.protocol,
+      hostname: u.hostname,
+      port: u.port || (isHttps ? 443 : 80),
+      method,
+      path: u.pathname + u.search,
+      headers,
+      timeout: 60_000
+    }, (res) => {
+      const chunks = []
```
duchnguyen, line 130: **[MEDIUM]** Response body buffered without size limit in hubFetch

All HTTP chunks are pushed into `chunks[]` and concatenated with `Buffer.concat` with no size cap. A `screen` call against a high-resolution device returns a base64-encoded PNG inside a JSON wrapper (Appium `/screenshot`), which can easily exceed 10 MB. The `cmdScreen` helper then calls `Buffer.from(b64, 'base64')` on top of the already-buffered response, doubling peak memory for that allocation. A rogue or misconfigured hub could also return an arbitrarily large payload for any endpoint.

skills/drive-automation-session/SKILL.md
```diff
@@ -0,0 +1,356 @@
...
+  --platformName "$PLATFORM_NAME" \
+  --udid "$UDID" \
+  --deviceName "$DEVICE_NAME" \
+  --platformVersion "$PLATFORM_VERSION" \
+  --automationName "$AUTOMATION_NAME" \
+  --app "$APP" \
+  --testingType "${TESTING_TYPE:-app}" \
+  --newCommandTimeout 1800 \
+  --scriptlessCapture \
+  > /tmp/drive-automation-session-caps.json
```
duchnguyen: **[MEDIUM]** Fixed /tmp caps path causes silent clobber under concurrent sessions

SKILL.md Steps 2–3 write rendered capabilities to the hardcoded path `/tmp/drive-automation-session-caps.json`, then read and `mv` it. If two agent sessions are started simultaneously (two terminals, two Claude conversations), the second write clobbers the first file before Step 3 reads it — the second session uses the wrong caps silently, connecting to the wrong device or app.

**Suggestion:**
```bash
# Step 2 — use a PID-qualified temp file
CAPS_TMP=$(mktemp /tmp/drive-automation-session-caps.XXXXXX.json)
node "$RENDER" ... > "$CAPS_TMP"

# Step 3 — read and move the unique file
SESSION_ID=$(node "$SKILL_DIR/scripts/appium.js" \
  --method POST --url /session \
  --req-body "@$CAPS_TMP" \
  | jq -r '.value.sessionId // .sessionId')
mv "$CAPS_TMP" "$SESSION_DIR/caps.json"
```

TranKienCuong:
Thanks @duchnguyen — all three addressed in `ed0b7c5`:

1. **`softFail` ignored `_base`** — `fail()` now takes an optional `explicitBase` that takes precedence over the `currentFlags`-derived path, and `softFail` forwards its `base`. So the error artifact is written correctly even if `softFail` is ever called before `currentFlags` is set.

2. **Unbounded response buffer in `hubFetch`** — added a 64 MB ceiling (`MAX_RESPONSE_BYTES`, env-overridable via `KOBITON_MAX_RESPONSE_BYTES` so the test can use a tiny limit). A high-DPI `/screenshot` stays well under it; a rogue/misconfigured hub now aborts instead of OOMing. The abort rejects the promise via `req.destroy()` with **no** error arg (passing an error there emits an unhandled `'error'` on the socket and crashes the process — caught that while testing), which flows through `main`'s catch → `softFail` → `iter-N.error.json`, exit 0. +1 vitest case.

3. **Fixed `/tmp` caps path clobbers under concurrent sessions** — switched Steps 2–3 to `mktemp`/`"$CAPS_TMP"` exactly as you suggested. This matters for the worktree workflow where parallel sessions are expected.

177/177 vitest; validate + codex mirror clean.

duchnguyen (approved):

huytunguyenn (approved):

PR #102, opened 2026-07-29
docs: per-skill host/platform/file-access requirements + compatibility matrix (KOB-54047, 1.9.0)

TranKienCuong:
## Why

Compatibility facts were real but scattered. `CLAUDE.md`'s Cross-tool surface table maps hosts → config — *where config lives*, not whether a skill can run at all. Each skill carried a free-text `compatibility:` field with inconsistent coverage, and the constraints that mattered most were buried mid-body: the CLI platform limit ~43 lines into `run-interactive-session`, the `Monitor` substitution ~169 lines into `monitor-test-run`. Nothing distinguished skills that need **local file access** from pure MCP glue that can run in a chat host with no filesystem.

So the question a user actually has — *"can I use this skill here?"* — had no single answer, and in one case the repo contradicted itself about it.

## What changed

**One compatibility matrix, in `CLAUDE.md`** (`### Skill compatibility matrix`, under Cross-tool surface): 5 skills × needs a persistent local filesystem · needs `~/.kobiton/.credentials` · local binary/OS constraint · needs a streamed-watch affordance · pure MCP. The existing host table is extended, not replaced.

Columns are keyed on **capabilities, not product names**. "Chat host" would have been the obvious shorthand for "no filesystem", but it's already wrong on some surfaces: a chat surface with code execution has a filesystem and Node, and *still* can't run the credential-dependent skills — because nothing there can execute `/automate:setup` to write `~/.kobiton/.credentials`. Keying on the capability keeps the matrix accurate as surfaces evolve, and makes the guidance actionable: "needs the credentials file `/automate:setup` writes" tells a user what to do next; "needs a CLI host" does not.

Each cell is checkable against that skill's `allowed-tools` **plus what its bundled scripts actually do** — both halves matter, because `allowed-tools` grants a superset. `create-test-run` declares `Read` yet bundles no files and reads none, so deriving from frontmatter alone would land on the wrong answer for its file-access cell.

`AGENTS.md` gets a prose `### Which skills run where` that **links** to the matrix instead of copying it — a table whose whole purpose is not drifting from reality shouldn't exist in two places.

**All 5 `SKILL.md`**: host, platform, and file-access requirements now open the existing `## Prerequisites`, with the route to take when unmet. One requirements list per skill, not a parallel block restating it. Each skill's `compatibility:` frontmatter was synced to the matrix too — otherwise a skill file can contradict itself, which is the exact failure this change exists to remove. (`validate.js` checks frontmatter structurally and can't catch a semantic contradiction, so that one needs a human pass.)

**`create-test-run` is the only pure-MCP, chat-safe skill.** Its Step 5 now confirms it can actually monitor before offering to — `monitor-test-run` runs a local poller, so on a filesystem-less host the right move is to create the run, hand back the id and portal link, and stop rather than offer a watch that can't happen.

## Corrections where the docs didn't match the artifact

**The bundled CLI is not "Apple Silicon only."**

```
$ file skills/run-interactive-session/bin/kobiton
… Mach-O 64-bit executable x86_64
$ lipo -info skills/run-interactive-session/bin/kobiton
Non-fat file: … is architecture: x86_64
```

Single-slice x86_64, no arm64 slice. It runs **natively on Intel Macs** — which the old claim wrongly excluded — and on Apple Silicon **only under Rosetta 2**, not natively as implied. Corrected across `CLAUDE.md`, `AGENTS.md`, and `README.md`.

The repo previously disagreed with itself here: the skill's own frontmatter said just "targets macOS" while the docs said Apple-Silicon-only. That contradiction is the clearest argument for having one matrix.

**The setup command** claimed binaries ship for "only macOS arm64 and Linux x64". `scripts/install-cli.sh` does no per-platform selection — it symlinks `run.sh` — no Linux binary ships, and the macOS binary is x86_64. Corrected in both `commands/automate/setup.toml` and `commands/setup.md`, which are meant to stay at parity.

**`drive-automation-session`'s `compatibility` field** described MCP `getCredential` with `~/.kobiton/.credentials` as a fallback. `scripts/appium.js` reads that file directly on every invocation and never calls `getCredential`, so the file is **required**. Its Errors table carried the same inversion and is corrected too.

## Where enforcement lives

Platform and credential checking stays in `scripts/run.sh`, which already validates the bundled binary (step 1) and the full credential set (step 4), reporting `/automate:setup` and `/automate:doctor` remedies. Skill docs state the constraint and route the user; they don't re-implement the check a layer up in markdown that may never execute.

An earlier revision of this branch embedded `uname`/`pkgutil` gates in the skill docs. They were removed during review, for two reasons: markdown can't enforce anything, and the duplication had already produced a real bug — a `pgrep oahd` Rosetta test that would have wrongly blocked a Rosetta-equipped Mac which simply hadn't run an x86_64 binary since boot. Better to state the fact once and let the code that actually runs do the checking.

## Version

Bumped to **1.9.0** — new capability, fully backward-compatible. `package.json` is the source of truth; the 6 host manifests were synced with `pnpm run build:version`. The 1.8.0 CHANGELOG entry keeps its original wording (including the superseded platform claim) since it's a historical record; the correction is noted in the 1.9.0 entry.

## Verification

Baseline captured **before** any edit, then re-run after:

| Check | Result |
|---|---|
| `pnpm run validate` | pass — skill frontmatter, `.codex/` mirror parity, version sync |
| `pnpm test` | **189/189**, 9 files — unchanged from baseline |
| Diff audit | no `tools/*.yaml`, no MCP server contract change |

`.codex/` regenerated with `pnpm run build:codex` — required by the `skills/**` edits, since `CLAUDE.md` and `AGENTS.md` aren't in the mirror's `PAIRS`. Never hand-edited.

## Follow-up worth its own change

`bin/kobiton` being x86_64-only means Apple Silicon users depend on Rosetta 2 being installed. Shipping a universal or arm64 slice would remove that dependency entirely. Real machine-checked platform enforcement, if wanted, belongs inside `run.sh` or the binary rather than in skill prose.


TranKienCuong:
### Code Review

Reviewed the full PR diff (23 files) with parallel agents at confidence threshold 80. Found **1 issue, now fixed** in 312ce6b.

**A precondition was stated more strictly than reality** — this change's own named failure mode.

`monitor-test-run` documents a foreground-loop fallback for hosts with no way to stream a background command's stdout: it degrades rather than refusing.

https://github.com/kobiton/automate/blob/312ce6b3a3531ef24f639116beae287c507c708a/skills/monitor-test-run/SKILL.md#L181-L185

But four surfaces described that affordance as **required**, and `create-test-run`'s Step 5 treated it as a hard third gate — refusing to offer monitoring at all without it. A host with a filesystem, credentials, and no streaming primitive can in fact be monitored, so the gate would have blocked a working setup.

Step 5 was also asking for a check it cannot perform: `create-test-run`'s `allowed-tools` is `Read, Skill` with no shell, so it can't probe for `~/.kobiton/.credentials`. It now judges only what it can actually know — whether it can run a local command at all — and hands off, letting `monitor-test-run` make its own credentials and streaming decisions.

https://github.com/kobiton/automate/blob/312ce6b3a3531ef24f639116beae287c507c708a/skills/create-test-run/SKILL.md#L141-L155

Aligned across the matrix cell, the `AGENTS.md` bullet, `monitor-test-run`'s frontmatter and Prerequisites, and `create-test-run`'s Prerequisites and Step 5.

https://github.com/kobiton/automate/blob/312ce6b3a3531ef24f639116beae287c507c708a/CLAUDE.md#L134-L136

#### Verified clean

- **Three-way consistency sweep** across all 5 skills — `compatibility:` frontmatter vs `## Prerequisites` intro vs matrix row — checked against the actual bundled scripts (`poll-test-run.js`, `appium.js`, `run.sh`). No disagreements on local-FS, credentials-file, OS, or streaming requirements.
- `run-automation-suite`'s "no credentials file" claim confirmed: neither its scripts nor `appium.ejs` reads that file or injects auth.
- `create-test-run`'s pure-MCP claim confirmed: no `scripts/` directory, no file reads.
- Version bump consistent at 1.9.0 across `package.json`, all 6 host manifests, and the CHANGELOG top entry.
- `.codex/` mirror in sync; `commands/setup.md` and `commands/automate/setup.toml` at parity.
- No stale prose describing the matrix's previous column set.
- `pnpm run validate` and `pnpm test` (189/189) pass, unchanged from the pre-change baseline.

Agents run: claude-md-auditor (0 findings), bug-detector (1). Security, DB-performance, history, and spec agents were skipped — a markdown-only diff with no code, SQL, migrations, or credential-handling logic gives them nothing to analyze.

---

huytunguyenn (approved):

skills/drive-automation-session/SKILL.md
```diff
@@ -347,7 +353,7 @@ The only hard programmatic stop is `MAX_ITERS=100` (override per session), which
 
 | Symptom | Likely cause | Fix |
 |---------|--------------|-----|
-| `No credentials available...` | Kobiton MCP not connected AND no `~/.kobiton/.credentials` | Either authenticate the Kobiton MCP server, or run `/automate:setup` |
+| `No credentials available...` | No `~/.kobiton/.credentials` (an authenticated MCP connection does not substitute — `appium.js` only reads the file) | Run `/automate:setup`, which uses the MCP connection to fetch and write the file |
 | `iter-N.error.json` has `status: 401` on session create | Credentials stale | Re-authenticate the MCP server (or re-run `/automate:setup` for the file fallback); check the portal URL |
```
duchnguyen: **[MEDIUM]** Stale troubleshooting row contradicts new credentials framing

The 401 row still says **'Re-authenticate the MCP server (or re-run `/automate:setup` for the file fallback)'**. The PR's own prerequisite text (line 61) now correctly states that `appium.js` reads `~/.kobiton/.credentials` directly and **never** calls the MCP `getCredential` tool — making 'Re-authenticate the MCP server' an ineffective remedy for a stale credentials file, and 'file fallback' a term the PR explicitly retired. An operator following this row would chase MCP re-auth and still see 401s until they re-run `/automate:setup`.

**Suggestion:**
```markdown
| `iter-N.error.json` has `status: 401` on session create | Credentials stale | Re-run `/automate:setup` to refresh `~/.kobiton/.credentials`; check the portal URL |
```

TranKienCuong: Good catch — applied verbatim in 50d7698.

You're right that it's not just wording: re-authenticating MCP genuinely doesn't fix a stale credentials file, since `appium.js` only ever reads `~/.kobiton/.credentials`. So the old row sent an operator down a path that could never clear the 401. I'd fixed the `No credentials available...` row two lines above and missed this one.

Also swept the repo for the same class of stale remedy. The only other hits are `CHANGELOG.md` (where "as a fallback" correctly describes the claim being corrected) and `monitor-test-run`'s equivalent row, which already pointed at `/automate:setup`.

PR #103, opened 2026-08-04
KOB-54335: replace the app tools with a unified promptApp

TranKienCuong:
Replaces the two app tools on this branch with a single **`promptApp`**, and carries `startNativeSession` across from KOB-54362 so the published catalog keeps it.

> **Scope changed mid-review.** This PR originally added `listAyaApps` + `askYourApp`. The ticket was rewritten to one unified tool and the branch was reworked in `8f151407`.

## One tool, no action parameter

`promptApp` takes a plain-language `question` and nothing else is required. The server routes the request itself — listing apps, app detail, exploring an app, explore status, and questions about an explored app's flows all go through the same tool. There is no action enum to pick from, because the backend decides how many of its own tools to run behind a single call.

The description carries the three things a model can't infer from the schema:

- **Conversation threading** — pass `conversationId` back to continue; omit it to start fresh.
- **`conversationStarted: true` means the previous conversation was not found**, so earlier context is gone. Without this a model would keep referring to a thread the server has already dropped.
- **`proposedAction` means a human must approve** — surface the link rather than trying to continue. Exploring an app spends real device time, so it is deliberately gated.

Selector lookups are called out explicitly ("what is the selector for the Add to cart button?"), since that is the main draw for automation work and not obvious from a tool named `promptApp`.

## Naming

No product name appears in the tool or its parameters — the underlying integration may still be renamed, and a tool name is a public contract that is painful to change later.

## `startNativeSession`

This branch is cut from `aya-develop`, which predates the `sessions.yaml` change on KOB-54362. Building the published catalog from this branch alone would have **silently dropped** that tool: it is registered server-side and live in the published catalog, and a handler with no schema entry simply stops appearing in clients' tool lists with no error anywhere.

The two branches touch disjoint files — that one owns `sessions.yaml`, this one owns `aya.yaml` — so taking it wholesale gives the correct union. Caught by diffing the built catalog against the **published** one, not just against this branch.

## Verification

`pnpm run validate` passes (`tools/aya.yaml is valid (2 tools)`, `tools/sessions.yaml is valid (6 tools)`, `.codex/` mirror in sync), `pnpm test` passes, and `pnpm run build:tools` emits a catalog whose only delta against the currently published one is `+promptApp / −listAyaApps / −askYourApp`.

Schema-only, as always in this repo — the handlers live server-side, and the names here match the registered handlers exactly.

### DIFF
--- a/tools/aya.yaml
+++ b/tools/aya.yaml
@@ -29,3 +29,77 @@ tools:
             logging.
       required:
         - userIntent
+
+  - name: listAyaApps
+    title: List Apps Available in AYA
+    description: >
+      List every app the user can ask AYA about: their own uploaded apps first,
+      then AYA's public app catalog. Call this before any other AYA app tool to
+      obtain an appId. The same bundleId may appear twice - once as the user's
+      private copy (isPublic false) and once as the public catalog copy
+      (isPublic true); these are separate apps with separate crawl data, so
+      state which one you mean. An app with crawlCount 0 or a null
+      firstCrawledAt has not been crawled yet and cannot answer questions.
+    annotations:
+      readOnlyHint: true
+      destructiveHint: false
+      openWorldHint: false
+    inputSchema:
+      type: object
+      properties:
+        q:
+          type: string
+          description: >-
+            Optional case-insensitive substring filter on app name and bundle
+            id. Exact substring only, not fuzzy - prefer omitting it and
+            filtering the results when unsure of the exact name.
+        limit:
+          type: integer
+          description: Maximum apps to return, 1-200, default 50.
+        userIntent:
+          type: string
+          description: >-
+            A brief summary of what the user is trying to accomplish with this
+            tool call, refined from their original prompt. Used for audit
+            logging.
+      required:
+        - userIntent
+
+  - name: askYourApp
+    title: Ask Your App
+    description: >
+      Answer a natural-language question about an app AYA has already crawled,
+      using its screen and flow knowledge graph. Identify the app by its AYA
+      appId from listAyaApps - never by bundle id, because one bundle id can
+      refer to both the user's private copy and the public catalog copy. Pass
+      conversationId from a previous result to ask a follow-up in the same
+      conversation; omit it to start a new one. Always pass the appId, including
+      on follow-ups. Returns an error if the app has not been crawled yet.
+    annotations:
+      readOnlyHint: true
+      destructiveHint: false
+      openWorldHint: false
+    inputSchema:
+      type: object
+      properties:
+        appId:
+          type: string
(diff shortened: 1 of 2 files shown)


TranKienCuong:
**Note on CI:** `.github/workflows/ci.yml` triggers only on `pull_request` to `main`, so no CI runs on a PR targeting `aya-develop`. Recording the local run here instead, from the branch tip:

```
pnpm run validate
  OK:   tools/aya.yaml is valid (3 tools)
  All checks passed.
  .codex/ mirror is in sync with source.
  Plugin version 1.8.0 is in sync across all manifests and CHANGELOG.md.

pnpm test
  Test Files  9 passed (9)
       Tests  189 passed (189)

pnpm run build:tools
  dist/tool-definitions.yaml contains: getAyaIdentityAssertion, listAyaApps, askYourApp
```

Worth deciding separately whether `aya-develop` should be added to the CI trigger list while it's an active integration branch.

d4rkwinz (approved):

d4rkwinz:
Post-merge notes from the AYA side (this landed on `aya-develop`, so flagging rather than blocking). All counts below are measured on **`aya-develop`**, not `main`.

### 1. README tool count is now stale

`README.md:393` says **"29 MCP tools"**. Actual on `aya-develop` is **33**:

| file | tools |
|---|---|
| apps.yaml | 5 |
| aya.yaml | 2 |
| devices.yaml | 4 |
| sessions.yaml | 6 |
| test-management.yaml | 14 |
| user.yaml | 2 |
| **total** | **33** |

The README also has **0 mentions of `promptApp`**, so the tool this PR is named for isn't documented yet.

### 2. "Replace the app tools" — but nothing was removed

The title says *replace the app tools with a unified promptApp*, and the diff is **+166/-0**. `apps.yaml` still ships its 5 tools alongside the new `promptApp`.

If that's deliberate staging (add now, remove in a follow-up), all good — worth saying so in the PR or a tracking issue. If not, the overlap is the kind of thing that degrades tool selection: a model facing both `promptApp` and five narrower app tools for the same job will pick inconsistently, which is harder to debug than either option alone.

### 3. No CI ran on this branch

`gh pr checks 103` reports **"no checks reported on the 'KOB-54335' branch"**. If there's a README/tool parity gate, it didn't execute here — which is why (1) got through. Worth confirming the gate runs on `aya-develop` PRs and not just `main`.

---

For cross-reference, the AYA-side contract this consumes: `stop_reason` is `proposal_pending` and `proposed_action` carries `{id, kind, display, approval_url, expires_at}` with `kind` ∈ `recrawl_app | run_test`. I've left the same note on kobiton/api#5077, whose fixtures currently use `awaiting_approval` and `{type: 'explore'}`.

PR #104, opened 2026-08-05
docs: correct Cursor UI labels and add the missing Cursor API key note

tungmhoang:
Three wrong UI labels and one missing client note, all found while writing the Cursor setup guides for the docs site ([kobiton/docs#547](https://github.com/kobiton/docs/pull/547), [KOB-53278](https://kobiton.atlassian.net/browse/KOB-53278)).

Everything here was verified against live Cursor CLI and Cursor IDE installs, and against the installed plugin on disk.

## UI label corrections

| Line | Was | Now |
|---|---|---|
| L166 | Cursor IDE: click **Connect** | **Authenticate** |
| L239 | Cursor CLI: select **kobiton** | **Kobiton** |
| L240 | Cursor IDE: click **Connect** | **Authenticate** |

L153 already said **Kobiton** and is unchanged, which is what flagged the inconsistency in the first place.

Worth noting the server name genuinely differs by surface: **Kobiton** in the Cursor CLI `/mcp list`, lowercase **kobiton** in the Cursor IDE **Tool & MCPs** panel. Both are left as they are, since both are correct.

## Missing Cursor entry under API Key Authentication

The note carries per-client guidance for Gemini CLI and Codex CLI but had nothing for Cursor, even though Cursor needs a different file *and* different variable syntax than any other client:

- Cursor reads `.cursor/mcp.json`, declared as `mcpServers` in `.cursor-plugin/plugin.json`. Not the root `.mcp.json`, which is the Claude Code config and sends `X-AI-Tool-Name: Claude`.
- Cursor expands `${env:NAME}`. Every other client uses a bare `${NAME}`.

That second point is the costly one. Copying another client's config produces no error. Cursor sends the literal string `${KOBITON_AUTH}` as the `Authorization` header and auth just fails.

The repo already ships the right template at `.cursor/mcp.apikey-example.json`; the note now points at it, and covers the plugin-cache versus project-level placement tradeoff.

## Note for reviewers

I only have READ on this repo, so this comes from a fork. Happy to hand the patch over if you would rather land it directly.

[KOB-53278]: https://kobiton.atlassian.net/browse/KOB-53278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,8 @@ For CI/CD pipelines or headless environments that cannot open a browser, use API
 
 > **Note:** OAuth and API key auth cannot coexist in a single `.mcp.json` (the API key config sets an `Authorization` header that OAuth must not have). To switch, replace `.mcp.json` with the appropriate format from `.mcp.apikey-example.json`.
 >
+> **`/automate:setup` does not work under API key auth.** It fetches your credentials through an *OAuth-authenticated* MCP session to write `~/.kobiton/.credentials`. Per the [skill compatibility matrix](CLAUDE.md#skill-compatibility-matrix), three skills read that file directly and have no MCP fallback, so they are unavailable on API key auth: `run-interactive-session`, `drive-automation-session`, and `monitor-test-run`. The MCP tools, `run-automation-suite` (your script carries its own credentials), and `create-test-run` (pure MCP) are unaffected.
+>
 > **Gemini CLI:** API key auth requires editing `gemini-extension.json` instead of `.mcp.json`. Add a `headers` block under `mcpServers.kobiton` with `"Authorization": "${KOBITON_AUTH}"`.
 >
 > **Cursor (CLI and IDE):** API key auth requires editing `.cursor/mcp.json` instead of `.mcp.json` - that is the file the plugin manifest points `mcpServers` at. Cursor expands `${env:NAME}`, not the bare `${NAME}` other clients use, so start from `.cursor/mcp.apikey-example.json` rather than another client's config:
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ The Cursor desktop editor installs the plugin from its built-in plugin browser:
 1. Open **Cursor Settings** > **Plugins** and paste `https://github.com/kobiton/automate` into the search box
 2. Click the **automate** result, then **Add to Cursor**, then **Install**
 
-To authenticate with the Kobiton MCP server: open **Tool & MCPs**, search for **kobiton**, click **Connect**, and complete the OAuth login in the browser.
+To authenticate with the Kobiton MCP server: open **Tool & MCPs**, search for **kobiton**, click **Authenticate**, and complete the OAuth login in the browser.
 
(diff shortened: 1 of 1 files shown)


tungmhoang:
Pushed 98fd937 after the paired docs PR ([kobiton/docs#547](https://github.com/kobiton/docs/pull/547)) surfaced a gap in my own wording here.

The Cursor bullet said to copy `.cursor/mcp.apikey-example.json` without saying where it lives. That reads fine if you cloned the repo, but readers who *installed* the plugin have that file buried at `~/.cursor/plugins/cache/kobiton/automate/<sha>/.cursor/`, which nobody finds by guessing. The identical wording was in the docs page and a reviewer went looking for the file, could not find it, and asked whether it shipped at all. It does; the instruction just was not actionable.

The bullet now carries the fetch command, mirroring the pattern already used for the project-only `.cursor/mcp.json` setup at L173:

```bash
mkdir -p .cursor
curl -sL -o .cursor/mcp.json https://raw.githubusercontent.com/kobiton/automate/main/.cursor/mcp.apikey-example.json
```

Still 1 file, README only, now +12/-4.

One thing I could not resolve from any source, flagged in case a maintainer knows: **does `/setup` work under API key auth?** The Getting Started section says setup uses the already-authenticated MCP session (OAuth) to fetch credentials. If an API-key session satisfies that too, both this README and the docs page could say so; right now both are silent and an API-key user cannot tell whether the step applies to them.

tungmhoang:
Answered: **`/automate:setup` does not work under API key auth.** Confirmed by testing.

That has wider consequences than the Cursor bullet, so `c3197d4` puts it in the **general** API Key Authentication note rather than a per-client one. Setup fetches credentials through an *OAuth-authenticated* MCP session to write `~/.kobiton/.credentials`, which it cannot do when auth is a static `Authorization` header.

Cross-referencing the [skill compatibility matrix](https://github.com/kobiton/automate/blob/main/CLAUDE.md#skill-compatibility-matrix), three of the five skills read that file directly with no MCP fallback, so API key auth takes them out:

| Skill | Under API key auth |
|---|---|
| `run-interactive-session` | unavailable |
| `drive-automation-session` | unavailable |
| `monitor-test-run` | unavailable |
| `run-automation-suite` | works, the user's script carries its own credentials |
| `create-test-run` | works, pure MCP |
| MCP tools | work |

Worth a maintainer sanity check, since I derived the per-skill breakdown from the matrix rather than testing each one. The `/automate:setup` failure itself is directly confirmed.

**This is a scope extension** beyond the Cursor fixes the PR started as. Happy to split it into its own PR if you would rather keep this one narrow, or drop it if the limitation is already understood and intentionally undocumented.

Now 3 commits, README only, +14/-4. The paired docs page ([kobiton/docs#547](https://github.com/kobiton/docs/pull/547)) carries the same information.

huytunguyenn (approved):

PR #105, opened 2026-08-05
docs: document the getOrgSettings tool and add a Documentation Parity rule

TranKienCuong:
## Summary

`getOrgSettings` shipped in `tools/user.yaml` but never made it into the published docs — the README Tools table, the `AGENTS.md` domain summary, and the `CLAUDE.md` tool inventory all still listed only `getCredential` for the user domain, so a reader could not discover the tool. This PR closes that gap, adds a contributing-guide rule for doc surfaces, and turns the tool/README slice of that rule into a CI gate in `validate.js`.

## Changes

**Docs**
- **README**: add `getOrgSettings` to the Account table; tool count 29 → 30
- **AGENTS.md**: rename the *Setup* bullet to *Account* (matching the README domain name) and list `getOrgSettings` alongside `getCredential`
- **CLAUDE.md**: add `getOrgSettings` to the tool inventory; drop `user.yaml` from the older-2-hint-shape notes — both its tools now carry the full annotation block, only `test-management.yaml` remains on the old shape
- **docs/examples.md**: stop claiming it covers "every tool" (it documents the common ones) and link the README for the full inventory
- **CONTRIBUTING**: new *Documentation Parity* section — a table mapping each change type (tool / skill / command) to the doc surfaces that must be updated in the same PR, with a pointer from the Pull Request Process checklist

**CI gate**
- **scripts/validate.js**: every tool name in `tools/*.yaml` must appear backtick-wrapped in `README.md` (backtick matching so `getApp` doesn't false-match inside `getAppParsingStatus`), and the README's "N MCP tools" count must equal the YAML total. Either drift now fails `pnpm run validate`, which CI runs on every push/PR — this is the check that would have caught `getOrgSettings` shipping undocumented. Skills and commands remain prose-only surfaces covered by the CONTRIBUTING table.
- **scripts/validate.test.js**: fixture gains a minimal README; 4 new tests cover the missing-README, undocumented-tool, stale-count, and prefix-name paths

## Testing

- `pnpm run validate` — all checks pass, including the new `README.md documents all 30 tools`
- `pnpm test` — 193/193 pass (189 existing + 4 new)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,7 +10,7 @@ Kobiton is a real-device mobile cloud for Android + iOS testing. This MCP plugin
 - **Apps**: list, upload, confirm upload, get parsing status, get details
 - **Sessions**: list, get, get artifacts, get user-input events, terminate
 - **Test management**: create / list / get / update / delete test cases, test runs, and test suites; `saveTestCase` converts a finished manual session into a reusable test case
-- **Setup**: `getCredential` (used by `/automate:setup`)
+- **Account**: `getCredential` (used by `/automate:setup`); `getOrgSettings` for org-level feature flags such as live remediation
 
 The MCP server runs at `https://api.kobiton.com/mcp`. Authentication is OAuth 2.1 (default) or API key (CI/headless).
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -57,7 +57,7 @@ CI runs `pnpm install --frozen-lockfile && pnpm run validate && pnpm test` on ev
 
 | File | Covers |
 |---|---|
-| `scripts/validate.test.js` | structural validation across manifests, tool YAMLs, skill frontmatter |
+| `scripts/validate.test.js` | structural validation across manifests, tool YAMLs, skill frontmatter, README tool parity |
 | `scripts/build-tool-definitions.test.js` | tool-definition YAML concatenation |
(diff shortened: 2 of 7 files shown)


TranKienCuong:
### Code Review

Found 1 issue:

1. **Prefix-matching test is not isolated** (test correctness, confidence 85/100)

The `does not match a tool name that is a prefix of a documented tool` test overwrites `devices.yaml` with 2 tools, but `setupValidProject`'s other 3 YAML files still contribute a `testTool` each — 5 tools total vs the fixture README's "4 MCP tools". The count-mismatch error fires alongside the intended missing-tool error, and `toContainEqual` tolerates it, so the test passes without isolating the prefix-matching path. Fix: have the test also write a README with the matching count (5), or assert the count-mismatch message is absent.

https://github.com/kobiton/automate/blob/bce8c8950bfd93a403688940ce84339bfaaba2c1/scripts/validate.test.js#L221-L236

Below-threshold notes (not blocking): the README presence check is a whole-file substring match, so a backticked mention in prose outside the `## Tools` table would satisfy it (78/100); `toolNames` is not deduplicated across YAML files, so a future duplicate tool name would produce a confusing count-mismatch failure (65/100).

---

TranKienCuong:
Review issue 1 addressed in d6554ad — the prefix test now writes its own README with the matching tool count and asserts the count-mismatch error is absent, isolating the prefix-matching path.

PR #112, opened 2026-08-10
KOB-54477: refresh bundled CLI so rejected file transfers exit non-zero (1.9.1)

TranKienCuong:
## What

Replaces the bundled `kobiton` CLI binary (both `skills/run-interactive-session/bin/kobiton` and the `.codex/` mirror) with a build from current CLI source (2026-08-07), and bumps the plugin to **1.9.1** across all manifests.

## Why

The previous binary (built 2026-07-21) swallowed the trailing gRPC status after end-of-stream on resource transfers: a `file push` the device side **refused** still printed `Pushed N bytes` and exited 0. The current CLI source propagates that status — a refused transfer now prints the server's reason and exits non-zero:

```
$ kobiton file push local.txt /system/blocked.txt
Server error: Tonic: status: Internal, message: "Tunnel closed Path is outside the safe directories: '/system/blocked.txt'."
$ echo $?
1
```

Happy-path pushes are unchanged (verified: push + file-list + exit 0).

## Notes

- Binary remains **macOS x86_64** (Rosetta 2 on Apple Silicon), matching the wrapper's documented contract; unsigned, same as the binary it replaces. The official release pipeline currently ships macOS **arm64** only — if we'd rather bundle the signed arm64 artifact, that needs the SKILL.md platform note updated and drops Intel-Mac support; happy to switch if preferred.
- One version bump only (1.9.0 → 1.9.1), applied to all seven manifests.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -11,7 +11,7 @@
   "plugins": [
     {
       "name": "automate",
-      "version": "1.9.0",
+      "version": "1.9.1",
       "description": "Connect to Kobiton mobile testing platform - manage devices, run automation suites, and view test results",
       "author": {
         "name": "Kobiton Inc.",
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "automate",
-  "version": "1.9.0",
+  "version": "1.9.1",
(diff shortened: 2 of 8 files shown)


TranKienCuong:
Folded into #111, which was already waiting on the CLI binary refresh — same binaries, minus the version bump (that branch follows the Unreleased-section convention; the release cut will name the version). Deleting this branch.

PR #116, opened 2026-08-13
feat: add startNativeSession tool for native automation sessions (KOB-54362, 1.11.0)

TranKienCuong:
## Summary

Adds `startNativeSession` to the Sessions domain — start a native automation session (XCUITest on iOS, UIAutomator on Android, or GameDriver) on a Kobiton device directly from an AI agent, without hand-crafted JSON-RPC.

## Changes

- `tools/sessions.yaml`: new `startNativeSession` tool definition. Only `userIntent` and `testFramework` are required; device targeting accepts an exact `udid` or `deviceName`/`platformVersion` matching; the app under test and test-runner bundle are passed as store identifiers from `uploadAppToStore`. Annotations follow the resource-creating pattern (`readOnlyHint/destructiveHint/idempotentHint/openWorldHint` all false).
- `README.md`: tool count 31; row added to the Sessions table.
- Version 1.11.0 (`package.json` source of truth; manifests synced via `pnpm run build:version`; CHANGELOG entry added).

## Notes

- The tool starts a real, billable device session; the description steers agents to `listDevices` first.
- `sessionTimeout`/`testTimeout` are in minutes; `deviceWaitTimeout` is in seconds and only honored for queued executions.
- Rebased on main after 1.10.0 shipped; validate + build + vitest (193/193) green.

---

### DIFF
--- a/tools/sessions.yaml
+++ b/tools/sessions.yaml
@@ -174,3 +174,117 @@ tools:
       required:
         - userIntent
         - sessionId
+  - name: startNativeSession
+    title: Start Native Session
+    description: >
+      Start a native automation session on a Kobiton device using a native test
+      framework — XCUITest (iOS), UIAutomator (Android), or GameDriver. Upload
+      the app under test and the test-runner bundle to Kobiton Store first
+      (uploadAppToStore) and pass their kobiton-store identifiers as `app` and
+      `testRunner`. Returns the created session ID plus the booked device UDID
+      and a test-report download URL. This starts a real, billable device
+      session — prefer listDevices first to confirm a matching device is
+      available.
+    annotations:
+      readOnlyHint: false
+      destructiveHint: false
+      idempotentHint: false
+      openWorldHint: false
+    inputSchema:
+      type: object
+      properties:
+        userIntent:
+          type: string
+          description: >-
+            A brief summary of what the user is trying to accomplish with this
+            tool call, refined from their original prompt. Used for audit
+            logging.
+        testFramework:
+          type: string
+          enum:
+            - XCUITEST
+            - UIAUTOMATOR
+            - GAMEDRIVER
+          description: >-
+            Native test framework to run. Exact upper-case values required.
+            XCUITEST is iOS-only; UIAUTOMATOR is Android-only.
+        sessionName:
+          type: string
+          description: Optional name for the session
+        sessionDescription:
+          type: string
+          description: Optional description for the session
+        deviceName:
+          type: string
+          description: >-
+            Device name to match (e.g. "Galaxy S21"). Ignored when `udid` is
+            given.
+        udid:
+          type: string
+          description: >-
+            Exact device UDID to run on. Takes precedence over deviceName /
+            platformVersion matching.
+        platformVersion:
+          type: string
+          description: OS version to match (e.g. "14", "17.4")
+        deviceGroup:
+          type: string
+          enum:
+            - KOBITON
+            - ORGANIZATION
+            - CLOUD
+          description: >-
+            Which device pool to book from. KOBITON = Kobiton cloud devices,
+            ORGANIZATION = your organization's own devices.
+        groupId:
+          type: integer
+          description: Team ID to run the session under
+        app:
+          type: string
+          description: >-
+            App under test, as a kobiton-store:<appId> identifier returned by
+            uploadAppToStore
+        testRunner:
+          type: string
+          description: >-
+            Test runner bundle (XCUITest test-runner .ipa, UIAutomator test
+            .apk, or GameDriver runner), as a kobiton-store:<appId> identifier
+        noReset:
+          type: boolean
+          description: Keep app data between runs (do not reset the app)
+        fullReset:
+          type: boolean
+          description: Uninstall and reinstall the app before the run
+        sessionTimeout:
+          type: integer
+          description: Maximum session duration in minutes (default 60)
+        testTimeout:
+          type: integer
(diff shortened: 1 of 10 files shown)


tools/sessions.yaml
```diff
@@ -174,3 +174,117 @@ tools:
...
+        udid:
+          type: string
+          description: >-
+            Exact device UDID to run on. Takes precedence over deviceName /
+            platformVersion matching.
+        platformVersion:
+          type: string
+          description: OS version to match (e.g. "14", "17.4")
+        deviceGroup:
+          type: string
```
duchnguyen, line 231: **[LOW]** `CLOUD` enum value in `deviceGroup` is undocumented

The `deviceGroup` description explains `KOBITON` and `ORGANIZATION` but omits `CLOUD`, leaving it ambiguous — especially since `KOBITON` is already described as "Kobiton cloud devices". An AI agent picking between the two cloud-sounding values will have no basis for the choice.

tools/sessions.yaml
```diff
@@ -174,3 +174,117 @@ tools:
...
+          description: >-
+            App under test, as a kobiton-store:<appId> identifier returned by
+            uploadAppToStore
+        testRunner:
+          type: string
+          description: >-
+            Test runner bundle (XCUITest test-runner .ipa, UIAutomator test
+            .apk, or GameDriver runner), as a kobiton-store:<appId> identifier
+        noReset:
```
duchnguyen, line 256: **[MEDIUM]** `noReset` and `fullReset` are mutually exclusive but not constrained

`noReset: true` (keep app data) and `fullReset: true` (uninstall and reinstall) are contradictory; submitting both will either cause a server-side error or silently prefer one. JSON Schema supports `not: { required: ["fullReset"] }` inside a conditional, or at minimum the descriptions should call out the conflict explicitly so the AI host knows to enforce mutual exclusion at prompt time.

tools/sessions.yaml
```diff
@@ -174,3 +174,117 @@ tools:
...
+          description: Only book devices carrying all of these device tags
+          items:
+            type: string
+        deviceWaitTimeout:
+          type: integer
+          description: >-
+            How long (in seconds, 1-86400) to wait for a matching device to
+            become available before failing. Only honored for queued
+            executions.
+        runCommand:
```
duchnguyen, line 283: **[MEDIUM]** `runCommand` is GameDriver-only but no schema constraint enforces that

The description says `runCommand` is "GameDriver only", but the schema places no `if`/`then` guard on `testFramework`. An AI agent running XCUITEST or UIAUTOMATOR may include `runCommand` (the field name is not framework-specific), receive an opaque server rejection, and have no schema-level signal to diagnose the error. At minimum, the description should add "Ignored or rejected for XCUITEST and UIAUTOMATOR." Ideally, a JSON Schema conditional makes the restriction machine-readable.

**Suggestion:**
```yaml
        runCommand:
          type: string
          description: >-
            GameDriver only — command line used to launch the runner.
            Passing this field for XCUITEST or UIAUTOMATOR is rejected by the server.
```

tools/sessions.yaml
```diff
@@ -174,3 +174,117 @@ tools:
...
+          description: Extra arguments passed to the test runner
+          items:
+            type: string
+        env:
+          type: object
+          description: Environment variables to set for the test runner process
+          additionalProperties:
+            type: string
+      required:
+        - userIntent
```
duchnguyen, line 299: **[MEDIUM]** `app` and `testRunner` absent from `required`, causing silent AI-agent failures

The tool description tells callers to upload binaries first then pass their `kobiton-store:` identifiers as `app` and `testRunner`, yet neither field appears in `required`. An AI agent that only sees the schema will happily omit both and receive an opaque server error. At minimum, add a note like "if omitted the server rejects the call" to each field's description; ideally add them to `required` (or document any framework-specific exemption — e.g., if GameDriver embeds the runner).

tools/sessions.yaml
TranKienCuong: Better catch than flagged: `CLOUD` was never a valid value — the server's device-pool selection handles `KOBITON`, `ORGANIZATION`, and `ANY` (defaulting to ANY), and an unmatched value finds no devices at all. Fixed in `0ee4998`: the enum is now `KOBITON | ORGANIZATION | ANY` with all three documented, including the default.

tools/sessions.yaml
TranKienCuong: Documented as mutually exclusive on both fields in `0ee4998`. Went with description-level guidance rather than a JSON Schema conditional — MCP hosts vary in how much of JSON Schema they enforce, so the description is the reliable channel to the agent.

tools/sessions.yaml
TranKienCuong: Documented in `0ee4998` as "Ignored for XCUITEST and UIAUTOMATOR" — checked the server behaviour: the field is stripped by the per-framework configuration allow-list rather than rejected, so "ignored" is the accurate wording (the suggested "rejected by the server" would misdescribe it).

tools/sessions.yaml
TranKienCuong: Documented in `0ee4998`: both fields now state they are required for XCUITEST and UIAUTOMATOR and that the server rejects the call when omitted. Kept them out of the schema `required` list deliberately — server-side validation owns the framework-dependent rule, and a hard schema requirement would block GameDriver's runCommand-driven variants.

duchnguyen (approved):

Issue #121, opened 2026-08-19
[Bug] The plugin's bundled docs diverge from the website in specific, reproducible ways — and the plugin is what an agent loads into context


informramiz:
## Summary

The website documents the product. The plugin ships its own `SKILL.md` / `AGENTS.md` / `--help`, and that is what an agent loads into context. In the concrete cases below the two disagree, and the agent has the version that is missing or wrong.

## What is on the website and does not reach the agent

- `test run … uiautomator` — with a worked example. The bundled docs do not carry it, and we concluded twice from them that instrumentation had no path on this platform at all. It is arguably the most valuable capability in the product.
- `device forward` — address format, a `--mode mux|demux` option we never knew existed, and *"Transient network connection failures are retried for up to 5 minutes"*, which would explain a 4-minute hang we spent a session diagnosing — we have no capture tying that hang to a retry, but the sentence never reached us, because it is absent from the bundled docs the agent was reading.
- `file push` — full syntax.

## Where they actively disagree

Two concrete cases we hit:

- **Test-run naming.** The plugin's bundled `test run` help and its own worked example on the same page disagree about the framework argument — the contradiction reported in issue #106.
- **adb allowlist.** The plugin's bundled adb notes describe a different permitted-command set than the published `manual-testing/adb-commands` page, and neither matches what a CLI/MCP session actually allows (the divergence reported in issue #107): commands the page forbids succeed, and permitted commands are blocked by an undocumented per-argument allowlist.

In both cases the plugin's version is the one in an agent's context, and the agent has no signal that a second, differing source exists — so it plans confidently from the wrong text and the error surfaces as a failed metered session rather than as a doc bug.

## A related discoverability trap: one name, two features

The name **"test run"** resolves to two unrelated features that do not cross-reference each other: the docs use it for `createTestRun` (re-executing a recorded test case), and the CLI uses `test run` for instrumentation. A reader who reaches the recorded-case page has no signal the instrumentation feature exists — it reads coherently enough to look like the answer. This is part of why we twice concluded instrumentation did not exist on this platform: the name we searched for resolved to something else. A one-line cross-reference on each surface, naming what it is *not* and pointing at the other, is enough — a rename is more disruptive than it is worth.

## Ask

Generate the plugin's bundled docs from the same source as the website, or have `--help` link to the canonical page per command. Any process that permits drift will keep producing this class of bug, and the reader least able to notice it is the one you are optimising for.

TranKienCuong:
[...]

PR #127, opened 2026-08-26
docs: bundled-docs catch-up, session create --hide, CLI pin bump, bug-report issue form (KOB-54716, 1.12.0)

### LINKED ISSUE #122: [Bug] The documented session-type enumeration is missing a value the API actually returns
## Summary

The `kobiton/automate` plugin's bundled `AGENTS.md` (plugin v1.9.0, the version we evaluated) enumerates the session types:

> *"every session has a type: `AUTOMATION` (script- or agent-driven Appium), `CLI` (the bundled CLI wrapper), `MANUAL` (a human driving the portal live view). If a human opens the live view and interacts while an automation session runs, the session becomes `MIXED`"*

So the documented set is `AUTOMATION` / `CLI` / `MANUAL` / `MIXED`.

`getSession` on a completed instrumentation run returned:

```
"type": "UIAUTOMATOR"
```

which is on none of those lists. The session was otherwise fully populated — `test_report_url`, `log_url` and `video_url` all present — so this was a normal, complete session, not a degenerate one.

Anything filtering or branching on session type from the documented enumeration silently drops instrumentation runs.

## Ask

Correct the enumeration, and check it against the values the API can actually emit rather than against the values it was expected to emit. Same species as the framework-argument and `deviceName`-filter issues: the product states a confident wrong thing rather than failing.

TranKienCuong:
## Summary

Second wave of pilot feedback on the bundled docs (#121, #122), plus adoption of the CLI's `session create --hide` and a structured bug-report form. Documentation, CLI pin, issue-template and manifest changes only — one version bump to **1.12.0**.

## Changes

- **Session-type enumeration** (`AGENTS.md`, `CLAUDE.md`, README glossary) now lists the values `getSession` / `listSessions` actually return — `AUTO`, `UIAUTOMATOR`, `XCUITEST`, `CLI`, `MANUAL`, `MIXED` (+ rarer `SCRIPTLESS`, `GAMEDRIVER`) and the `automation_type` field — verified against live API responses. There is no `AUTOMATION` value. Fixes #122.
- **`run-interactive-session` docs** (`SKILL.md`, `references/response-shapes.md`): `test run --app --runner <uiautomator|xcuitest>` row + "Instrumentation runs" section with worked examples; `device forward --mode mux|demux` and the 5-minute transient-retry note; `session list` (+ `list-active`, `show`); `session create --hide` in every example with the token-handling notes; corrected the `session create` output shape (`Session <id> created for device <udid>.`). Closes the remaining items of #121.
- **Cross-reference** between the two features that share the name "test run": README `createTestRun` row and `create-test-run` skill description point at the CLI's instrumentation command and vice-versa.
- **CLI pin** `skills/run-interactive-session/CLI_VERSION`: `2608.70138.0+master.b0c2bc7` → `2608.191335.0+master.1e8f487` (first published build with `--hide`, `session list`, `forward --mode`; sha256 verified; validated with a live install → create `--hide` → ping → adb-shell → list → end flow).
- **Bug-report template**: `.github/ISSUE_TEMPLATE/bug_report.md` replaced by the issue form `bug_report.yml` — plugin version, AI host + version, OS/arch, tool/skill, repro, expected and actual are required; CLI version and `/automate:doctor` output optional.
- `package.json` 1.12.0 → manifests via `sync-version.js`; `.codex/` via `sync-codex-artifacts.js`; CHANGELOG entry.

## Test Plan

- [x] `pnpm run validate` (mirror + version lockstep) and `pnpm test` (193/193) pass
- [x] Fresh install downloads the pinned build; `session create --hide` prints no token line and `ping` / `adb-shell` / `session list` / `session end` work; the session reports `type: CLI`
- [x] Issue form parses; required ids = plugin-version, host, host-version, os, surface, repro, expected, actual
- [ ] Form renders in the "New issue" chooser after merge (GitHub reads templates from the default branch only)

## Checklist

- [x] Code follows project conventions
- [x] Documentation updated
- [x] No sensitive data exposed
- [x] CHANGELOG.md updated


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,111 @@
+name: Bug Report
```
huytunguyenn, line 1: This PR deletes `.github/ISSUE_TEMPLATE/bug_report.md`, but two README links still point at it:
- README.md:188: the "open an issue" link
- README.md:796: the support paragraph link

Both use `?template=bug_report.md`. GitHub resolves `?template=` by exact filename, and `config.yml` sets `blank_issues_enabled: false`, so clicking these now lands on the template chooser instead of the new bug form. 

Should change both to `?template=bug_report.yml`.

skills/run-interactive-session/references/response-shapes.md
```diff
@@ -19,21 +19,23 @@ Most WebDriver endpoints return a JSON envelope `{"value": <result>}`. A few com
...
-| `session ping` | Exit code 0 = alive, non-zero = expired | Trust exit status; don't parse stdout |
-| `session end` | Text confirmation | No parsing needed |
+| `session create --hide` | One line: `Session <id> created for device <udid>.`; the session token is **not** printed (it is still saved to `~/.kobiton/.session`) | `grep -oE 'Session [0-9]+ created'` and take the number. Without `--hide` a second line `Session token <jwt>` appears - always pass the flag, and never echo that line if you see it |
+| `session ping` | `Session <id> pinged.`; exit code 0 = alive, non-zero = expired | Trust exit status; don't parse stdout |
+| `session list` | Comma-separated table: header `ID, State, Type, Device, Platform, Created, Ended`, one session per row (`Ended` is `active` for running sessions), then a footer `Page N (M items), T total`; default 20 rows per page, `--all` returns every page in the window | Read directly or `grep` for an id / device name; narrow with `--state`, `--type`, `--platform`, `--keyword`, `--from` / `--to` rather than parsing a large dump. The `Type` column carries the same values as `getSession` (`CLI`, `AUTO`, `MANUAL`, `UIAUTOMATOR`, ...) |
```
huytunguyenn: Two `session list` output shapes are missing here, both on paths this cell recommends:

**1. `--all` prints a different footer.** Verified:

```
$ session list --from 2026-08-20 --to 2026-08-21 --all
...
352 of 352 sessions, 18 pages.
```

Not `Page N (M items), T total`. Since this cell tells the agent to use `--all`, parsing the documented footer returns nothing on exactly that path. Worth documenting both footers (paged vs `--all`).

**2. Empty results print a sentinel with no header row.** Verified with `--state START`:

```
No sessions found.
```

An agent grepping for the header/rows can't tell "no matching sessions" from a failed call. Since this file is the authoritative parsing recipe, the sentinel line belongs in it.

AGENTS.md
huytunguyenn, line 1: This file picked up the session-type correction, but not the `--hide` rule.

Gemini CLI doesn't load skills since`gemini-extension.json` points `contextFileName` at `AGENTS.md`, so this is effectively all the context it gets. Step 2 under "When the user asks to interactively drive a device" just says "start an interactive session", with no mention of `--hide`:

> 2. **Create or resume a session**: `reserveDevice` then start an interactive session; resume an existing one by session ID if the user has one.

An agent following that runs `session create` without the flag, and the bearer token gets printed into the transcript. Therefore, the leak this PR closes stays open on non-Claude hosts.

Not a guaranteed leak, to be fair: line 96 does point at `SKILL.md`, so an agent may read the rule there. But `CLAUDE.md`'s Cross-tool surface section asks for substantive skill-workflow changes to be mirrored here, and this file already mirrors comparable operational detail (the full restricted `adb-shell` policy at line 98).

Suggest adding one clause in step 2, such as "start an interactive session with `session create --hide` so the bearer token stays out of stdout". The same pass could mirror `session list`, `test run --runner`, and `forward --mode`.

skills/run-interactive-session/SKILL.md
```diff
@@ -127,15 +129,15 @@ Capture both the **UDID** (used for session creation) and the device **id** (the
...
-    $KOBITON_BIN -u <udid> session create
+    $KOBITON_BIN -u <udid> session create --hide
 
-The output contains a line like `kobitonSessionId: 12345`. Capture it:
+The output is a single line like `Session 12345 created for device <udid>.`. Capture it:
 
-1. Parse the session ID from the output.
+1. Parse the session ID from the output (`grep -oE 'Session [0-9]+ created' | grep -oE '[0-9]+'`).
 2. Create the artifacts directory: `mkdir -p .kobiton/sessions/<session-id>`.
 3. Store the session ID for use in screenshot and page source commands.
```
huytunguyenn, line 139: `--hide` is mandated here, but no minimum CLI build is stated — and `run.sh` can hand the agent a build that doesn't have the flag.

`run.sh` only resolves, never downloads (`run.sh:25`). If the pinned build isn't cached, `run.sh:34-45` falls back to the newest cached build. Two paths in `install-cli.sh` land there:

- **`rc=2`** (download failed, cache exists) — `install-cli.sh:233-235`, whose own comment reads *"offline but a cached build exists - run.sh will use it; stay quiet"*
- **`rc=4`** (pin pruned upstream, cache exists) — `install-cli.sh:209-210`, keeps the cached build

For existing users the cached build is the previous pin, `2608.70138.0+master.b0c2bc7` — which per this PR's CHANGELOG predates `--hide`.

**Concrete failure:** a user on 1.11.0 has `2608.70138.0` cached → updates to 1.12.0 → the first session runs offline or behind a proxy → `install-cli.sh` quietly keeps the old build → `session create --hide` fails as an unknown flag → the agent follows the recovery rule at lines 104 and 410 ("run `--help` and retry with the right arguments") → retries **without** `--hide` → the bearer JWT is printed into the transcript. That is the exact outcome this change exists to prevent.

The parse recipe just above (line 135) also matches nothing against the old build's `kobitonSessionId: <id>` output, so the agent can't capture the session id either.

Worth noting this isn't a narrow edge case — it needs an older cached build, which every upgrading user has. A fresh install with no cache gets a loud error instead, not a silent downgrade.

**Two ways out:**

1. State the minimum build for `--hide` (and `session list` / `forward --mode`) here, and tell the agent that the `pinned CLI version ... is not cached` warning means the flag may be missing.
2. Better: have `run.sh` inject `--hide` into `session create` when it's absent, so the guarantee lives in the wrapper instead of depending on which build resolved and on the agent not dropping the flag during error recovery.

Longer term this points at the CLI itself: if hiding the token were the default, with an opt-in `--show-token` for the rare script that needs the raw JWT, there would be no mandate to enforce and no version floor to document.

.github/ISSUE_TEMPLATE/bug_report.yml
TranKienCuong: Good catch - both README links now point at `?template=bug_report.yml` (3263729).

skills/run-interactive-session/references/response-shapes.md
TranKienCuong: Thanks, both verified against the pinned build and documented in 3263729: the paged footer `Page N (M items), T total`, the multi-page `--all` footer `N of T sessions, P pages.` (a single-page `--all` keeps the paged footer), and the `No sessions found.` sentinel with no header row, with the parsing recipe told to check for the sentinel first and accept either footer. One thing I noticed while verifying: on this build the multi-page footer printed `10 of 0 sessions, 3 pages.` - the total is wrong in the CLI itself, so the doc says to count rows rather than trust that number. I'll raise it against the CLI separately.

AGENTS.md
TranKienCuong: Agreed - AGENTS.md step 2 now spells out `kobiton -u <udid> session create --hide` (why, what it prints, where the token still lives) plus `session list`, and step 3 carries `device forward --mode mux|demux` with the 5-minute retry note and `test run --app --runner <uiautomator|xcuitest>` with the createTestRun distinction, so Gemini / Cursor / Copilot get the same rule without loading the skill (3263729).

skills/run-interactive-session/SKILL.md
TranKienCuong: This is the right call and I took option 2. `run.sh` now owns the guarantee (3263729): if the invocation is `session create` it checks the resolved build for `--hide` support via `session create --help`; when supported it appends `--hide` if the caller omitted it, and when the resolved build predates the flag (exactly the offline-upgrade / pruned-pin fallback you traced) it refuses to run the command and points at `/automate:setup` - failing closed instead of printing the token. Verified the four branches with stub binaries (inject / already present / non-create commands untouched / old build refused) and live on a device: `session create` without the flag through the wrapper produced only `Session <id> created for device <udid>.`. SKILL.md now states the minimum build (2608.191335.0), explains what the `pinned CLI version ... is not cached` warning implies for the flag, and the error-recovery rule has an explicit exception: never drop `--hide`, the refusal message means fetch the pinned build. Agree the long-term fix is hide-by-default with an opt-in `--show-token` in the CLI; noting it for the CLI side.

huytunguyenn (approved):
All six review findings are addressed, and the `--hide` one is fixed better than suggested — the wrapper now enforces the guarantee instead of the docs merely mandating it.

Verified by running, not just reading:

**`run.sh` enforcement**
- Arg detection is correct across realistic shapes: `session create`, trailing flags, a global flag with a value ahead of the subcommand (`-u <udid> session create`), and `--flag=value`. No double `--hide` when it's already passed, and no false positive on `device create` / `app create` / `session list|end|ping|show`.
- Refusal fails closed: against a stub build whose `session create --help` lacks `--hide`, it exits 1 on stderr and never execs the binary.
- The `--help` probe takes ~0.13s and needs no network or credentials.
- The `pipefail` + `grep -q` SIGPIPE race doesn't fire: help output is 780 B, well inside the 64 KB pipe buffer (0 false refusals in 200 runs).
- `session create` is the only command that prints the token, so the single enforcement point covers it.

**Docs**
- README: both links now `?template=bug_report.yml`.
- `SKILL.md`: version floor stated, the "retry with the right arguments" recovery rule now carries an explicit never-drop-`--hide` exception, and the refusal has its own error row — that recovery path was the actual leak vector.
- `AGENTS.md`: `--hide` plus `session list`, `forward --mode`, and `test run` mirrored for hosts that read it instead of the skill.
- `response-shapes.md`: both footers and the `No sessions found.` sentinel, and it goes further than my report — single-page `--all` keeps the paged footer, and `T` can read `0`.

`pnpm run validate` passes (`.codex/` mirror in sync, 1.12.0 lockstep); `pnpm test` is 193/193.

Two non-blocking notes:

1. The new `run.sh` logic has no test. The repo has no bash test harness (all 9 test files are vitest over JS), so there's no precedent — but this is a security enforcement path and worth covering later.
2. The refusal is a behavior change that can block a previously working flow: a user offline with an older cached build used to get a session (and a leaked token) and now gets a hard stop. That's the right call and the message says what to do — noting it so it's a known tradeoff rather than a surprise.

My earlier MINOR-vs-PATCH note no longer applies: this is no longer docs plus a pin refresh, it adds real wrapper behavior, so 1.12.0 fits.

LGTM.

PR #129, opened 2026-09-03
[KOB-54806][KOB-54809] Team-aware test-management descriptions, listTeams tool, sortBy enum

TranKienCuong:
## Summary

Team-scoping accuracy fixes for the test-management tools, plus one new discovery tool.

The test-management list tools require `teamId` for organization members (the API returns `400 Team id is required` without it), while organization admins omitting `teamId` get org-level (team-less) items only. The previous descriptions claimed org-wide listing "in addition to org-shared ones" — wrong for both roles — and an agent hitting the 400 had no tool to discover a valid `teamId`. Separately, `listTestRuns.sortBy` documented `createdAt` as its example, a value the API's `NAME`/`CREATED_AT` enum can never accept.

## Changes

- **New `listTeams` tool** — returns the calling user's teams (id, name) for recovering from `Team id is required` and for scoping team-based calls; an empty result means the caller belongs to no team.
- **Corrected `teamId` descriptions** on `listTestCases`, `listTestRuns`, `listTestSuites`, `createTestSuite`: admins omit ⇒ org-level items; members must pass one (discover via `listTeams`); results carry `team_id` (`null` = org-level).
- **`saveTestCase`** now documents that the created test case inherits the source session's team; **`createTestRun`** documents the team-homogeneity constraint.
- **`listTestRuns.sortBy`** gains `enum: [NAME, CREATED_AT]` with an accurate description.
- Touched tools extended with the standard `idempotentHint` / `openWorldHint` annotations per the repo convention; README tool table updated (32 tools).

Server-side behavior for `listTeams` and the `sortBy` normalization ships separately; the published tool definitions will be refreshed after that deployment.

## Test Plan

- [x] `pnpm run validate` (tool YAMLs, README parity, mirrors, version sync)
- [x] `pnpm test` (193/193)

---


tools/test-management.yaml
```diff
@@ -49,9 +54,11 @@ tools:
         teamId:
           type: integer
           description: >-
-            Optional team scope. When set, only test cases visible to the
-            caller via that team are returned (in addition to org-shared ones).
-            Omit to list every test case the caller can access in the org.
+            Team scope. Organization admins: optional - omit to get org-level
```
huytunguyenn: This parameter text now says `teamId` is **required** for org members. But the tool-level summary a few lines above (line 38) still says "with optional team ... filters", and the same wording is in `listTestRuns` (line 341), `listTestSuites` (line 449) and the README rows for those three tools.

Agents read the summary first. An org member who follows it will call without `teamId` and get 400 "Team id is required". An admin asking for "all our test cases" will omit it and get only team-less rows, then report a partial list as if it were complete.

Suggest updating the three summaries and README rows to match, e.g. "List test cases for one team (teamId required for org members; admins may omit it to get org-level items only)."

tools/test-management.yaml
```diff
@@ -9,10 +9,14 @@ tools:
       enforces that the session is a manual, prime-completed, revisit-eligible
       session owned by the user. The test case is named after the source
       session; call updateTestCase with action 'update_info' afterward to
-      rename, describe, or retag it. Returns the created test case record.
+      rename, describe, or retag it. Returns the created test case record. The
+      created test case inherits the source session's team (team_id in the
```
huytunguyenn, line 13: Worth a warning here: for an org member, a test case with `team_id = null` cannot be found later through `listTestCases` or `listTestSuites`, because `teamId` is mandatory for members and the service filters on that exact team.

This is reachable: a manual session booked without `groupId` is saved with `groupId = null`, so `saveTestCase` produces a team-less case. The member then lists with a team id from `listTeams`, the new case is missing, and the agent concludes the save failed and retries.

Suggest adding one sentence, e.g. "If the session had no team, the test case is org-level and is not visible to org members through listTestCases; use getTestCase with the returned id."

tools/test-management.yaml
```diff
@@ -366,7 +380,10 @@ tools:
           default: 10
         sortBy:
           type: string
-          description: Sort key (e.g. `createdAt`)
+          enum:
```
huytunguyenn, line 389: Small one: this enum is enforced case-sensitively by Ajv at the MCP hop (no coercion), while `GET /v2/test-runs` accepts any casing and upper-cases it. So `sortBy: "created_at"`, which worked on main and is valid on the server, now fails with "must be equal to one of the allowed values".

The enum is still a net gain over main's free text. Just note "exact upper-case values" in the description, and the server-side normalization mentioned in the PR body becomes unnecessary for MCP callers.

tools/test-management.yaml
```diff
@@ -513,7 +538,13 @@ tools:
             logging.
         teamId:
           type: integer
-          description: Optional team ID that owns the new test suite
+          description: >-
+            Team that owns the new test suite. Organization admins may omit it
```
huytunguyenn: The team-scoping paragraph now appears in five places: `listTestCases`, `listTestRuns`, `listTestSuites` (identical apart from the noun), this variant on `createTestSuite` (which drops the 400 text), and again inside `listTeams`. That adds roughly 2.5 KB of repeated text to the catalog every host loads per session, and drift is already visible in this PR.

Simpler shape: keep the full explanation once (in `listTeams` or a short "Team scoping" note in README) and put a one-liner on each parameter, e.g. "Team scope. Required for org members, optional for admins (omit = org-level only). Get ids from listTeams."

YAML anchors would only dedup the source file; `js-yaml` dump inlines string aliases, so the published bytes stay the same.

tools/test-management.yaml
```diff
@@ -601,3 +632,30 @@ tools:
       required:
         - userIntent
         - testSuiteId
+  - name: listTeams
```
huytunguyenn: Deploy-ordering risk. On `api` master, `registerHandlers` has no handler for `listTeams` yet (it lives on the KOB-54806 branch). The server re-downloads the S3 YAML about every 10 minutes and `tools/list` advertises every tool in the YAML whether or not a handler exists.

If `dist/tool-definitions.yaml` is published before the api change is deployed, within 10 minutes every host sees `listTeams`, the new `teamId` text steers members to call it, and the server answers JSON-RPC -32601 "Method not found: listTeams". That is a protocol error, not a tool `isError`, so the agent has no recovery path.

The PR body mentions the ordering, but nothing in this repo enforces it. At minimum please call it out in the release checklist for this version.

tools/test-management.yaml
```diff
@@ -601,3 +632,30 @@ tools:
...
+    title: List Teams
+    description: >
+      List the teams the calling user belongs to (id and name). Use this to
+      discover a valid teamId for the team-scoped test-management tools: an
+      organization member must pass teamId to listTestCases, listTestSuites
```
huytunguyenn: Two things about this guidance:

1. It is written reactively ("when one of those returns 'Team id is required', call this tool and retry"). Admins never hit that 400, so they are never steered here. The important fact is missing: **there is no cross-team listing at all**. The only org-wide route is `listTeams`, then one list call per team, plus the omit-`teamId` call for the null bucket. Without that sentence, an admin asking for "all test runs in the org" gets only team-less rows and no signal the answer is incomplete. No MCP tool exposes the caller's role, so the agent cannot know in advance whether omitting `teamId` will 400 or silently narrow.

2. The list of tools is incomplete. `createTestSuite` also returns the same 400 when a member omits `teamId` (`POST /v2/test-suites` runs `authorizeTcmV2` on `request.body.team_id`), and its own description already points back here. Either add it, or name no tools.

Suggested wording: "There is no cross-team listing. When the team is not already known, call listTeams first and query each team; org admins may omit teamId to get org-level (team-less) items only."

huytunguyenn (commented):
Automated review of the PR head. `pnpm run validate` and `pnpm test` (193/193) pass, so nothing below is a build break. All findings are about the published tool contract and its docs.

Inline comments cover the YAML. A few items land on lines outside the diff, so they go here:

**CLAUDE.md is now out of date (not touched by this PR)**
- The tool inventory still says `tools/test-management.yaml` holds "14 test-case / test-run / test-suite CRUD tools". After this PR it has 15, and `listTeams` is a lookup, not CRUD. README says 32 tools, CLAUDE.md sums to 31, and `validate.js` only checks README.
- Two places still say the file "uses the older 2-hint shape". After this PR the shape is mixed (see next point). CONTRIBUTING lists the CLAUDE.md inventory as something to update in the same PR that adds a tool.

**Annotation hints were extended on only 7 of 15 tools**
- `getTestCase`, `updateTestCase`, `deleteTestCase`, `getTestRun`, `terminateTestRun`, `getTestSuite`, `updateTestSuite`, `deleteTestSuite` are still on the 2-hint shape. `terminateTestRun` in particular is missing the `idempotentHint: true` the verb table in CLAUDE.md prescribes.
- CLAUDE.md asks to finish the migration when touching this file. The server ignores annotations (`tools-registry.js` only stores name / description / inputSchema), so doing it here has zero runtime risk.

**Minor / judgment call**
- `listTeams` takes only `userIntent` and returns caller identity data, the same shape as `getCredential` / `getOrgSettings` in `tools/user.yaml`. Filed under `TEST_MANAGEMENT`, every call is audited as `MCP_CALL_TEST_MANAGEMENT`, and its result also feeds `listDevices.groupId` and `startNativeSession.groupId`. Moving it to `user.yaml` (plus the README Account section) is cheap now, before the handler ships and rows accumulate.
- `AGENTS.md` has no note about `teamId` / `listTeams` yet, and the `create-test-run` skill error table has no row for the 400 "All test cases must have the same team". Both are soft; hosts still get the YAML text over MCP.

TranKienCuong:
Applied the listTeams → tools/user.yaml move (with the README Account-section row) in 9735e9f — agreed it belongs with the other caller-identity tools, and the audit domain follows. validate + 193/193 tests green. The remaining review items are under consideration and will be addressed separately.

TranKienCuong:
All remaining review items applied in e13fa2c + ddf0dab:
- The three list-tool summaries (and README rows) now state the team contract up front
- teamId parameter texts deduplicated to one-liners; the full explanation lives once in listTeams, which now leads with the no-cross-team-listing fact instead of the reactive 400 framing
- saveTestCase warns about the team-less/org-level visibility trap
- sortBy notes exact upper-case values
- Annotation-hint migration finished on the remaining 8 tools; CLAUDE.md inventory and the stale 2-hint prose updated
- AGENTS.md team-scoping note + create-test-run error row for the team-homogeneity 400

Thanks — every item landed.

huytunguyenn (approved):

